PR #116, opened 2026-07-02
fix(java-adapter): honor SKIP_ADAPTER_VENDOR and skip gracefully on old javac

debugmcpdev:
## Problem

The nightly **Flake Hunt** workflow has failed on every run since it was added in #110 — on both `ubuntu-latest` and `windows-latest` — before any tests execute:

```
packages/adapter-java build:adapter: error: release version 21 not supported
[compile-jdi-bridge] Compilation failed
ELIFECYCLE Command failed with exit code 1.
```

## Root cause

Flake Hunt intentionally installs no language toolchains and sets `SKIP_ADAPTER_VENDOR=true` on `pnpm install`. The rust and javascript vendor scripts honor that variable, but `compile-jdi-bridge.js` only honored its own `SKIP_JDI_COMPILE`, so the JDI bridge compile ran anyway. The hosted runners ship Temurin **17** as the default JDK (`/usr/lib/jvm/temurin-17-jdk-amd64`, `Java_Temurin-Hotspot_jdk\17.0.19-10`), and javac 17 cannot target `--release 21`, so the root postinstall hard-failed. Main CI never hits this because it runs `actions/setup-java` with JDK 21 before installing.

## Fix

- Skip JDI bridge compilation when `SKIP_ADAPTER_VENDOR=true`, matching the convention used by the rust/javascript vendor scripts (and documented in the README).
- When javac is present but older than JDK 21, warn and skip instead of failing the install — the same graceful degradation already used when javac is missing entirely ("adapter will report the error at runtime"). Genuine compile errors under JDK 21+ still fail hard, so main CI keeps its protection.

## Verification

All paths exercised by running the script directly on Windows:

- `SKIP_ADAPTER_VENDOR=true` → skip message, exit 0
- Normal run with JDK 21 → compiles successfully (javac under `C:\Program Files\...` — spaces handled via `execFileSync`)
- Too-old javac (simulated by temporarily raising the required release) → warns and exits 0
- Version regex handles `javac 17.0.19` → 17, `javac 21.0.10` → 21, `javac 25-ea` → 25, empty output → `null` (proceeds and lets javac itself report)

After merge, a `workflow_dispatch` run of Flake Hunt should confirm the fix end-to-end.

### DIFF
--- a/packages/adapter-java/scripts/compile-jdi-bridge.js
+++ b/packages/adapter-java/scripts/compile-jdi-bridge.js
@@ -20,6 +20,7 @@ const JAVA_DIR = resolve(__dirname, '..', 'java');
 const SOURCE_FILE = resolve(JAVA_DIR, 'JdiDapServer.java');
 const OUT_DIR = resolve(JAVA_DIR, 'out');
 const CLASS_FILE = resolve(OUT_DIR, 'JdiDapServer.class');
+const TARGET_RELEASE = 21;
 
 function findJavac() {
   // Check JAVA_HOME first
@@ -40,6 +41,20 @@ function findJavac() {
   return null;
 }
 
+function getJavacMajorVersion(javac) {
+  try {
+    // `javac -version` prints "javac <major>.<minor>.<patch>" on stdout (JDK 9+)
+    const output = execFileSync(javac, ['-version'], {
+      encoding: 'utf-8',
+      stdio: ['ignore', 'pipe', 'pipe']
+    });
+    const match = /javac\s+(\d+)/.exec(output);
+    return match ? Number(match[1]) : null;
+  } catch {
+    return null;
+  }
+}
+
 function needsCompilation() {
   if (!existsSync(CLASS_FILE)) return true;
   if (!existsSync(SOURCE_FILE)) return false; // nothing to compile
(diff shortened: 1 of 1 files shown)


debugmcpdev:
**Validation via `workflow_dispatch` on this branch**: [run 28575215007](https://github.com/debugmcp/mcp-debugger/actions/runs/28575215007)

✅ **The step this PR fixes now passes on both OSes** — `Install Node dependencies` (previously died in the JDI compile) and `Build project (CI)` are green on `ubuntu-latest` and `windows-latest`.

❌ The `Run flake hunt` step itself fails, but for a **separate, pre-existing reason** this PR does not attempt to fix: now that the hunt actually executes for the first time, it exposes unit tests that are not hermetic. Across 20 seeded runs (10 per OS), the failures are deterministic, not order-dependent:

| Test file | Failures | Likely cause |
|---|---|---|
| `packages/adapter-javascript/.../javascript-debug-adapter.command.test.ts` (5 tests) | 20/20 runs | needs vendored js-debug (skipped via `SKIP_ADAPTER_VENDOR`) |
| `packages/adapter-javascript/.../javascript-debug-adapter.command.edge.test.ts` (3 tests) | 20/20 runs | same |
| `packages/adapter-javascript/.../factory-export.test.ts` (1 test) | 20/20 runs | same |
| `tests/unit/shared/adapter-policy-go.test.ts` (`validateExecutable`) | 5/20 runs | spawns a real executable |
| `packages/adapter-rust/tests/rust-adapter.test.ts` (`should validate environment`) | 1/20 runs | probes real cargo/CodeLLDB; also flakes locally (15s timeout vs 155ms in isolation) |

Main CI doesn't hit these because it runs `pnpm run vendor:adapters` as an explicit step with toolchains installed. Per the workflow header ("If a unit test ever needs one, that is itself a hermeticity bug to fix"), these deserve their own follow-up.

Issue #122, opened 2026-07-05
Windows: process tree leaks — dev-proxy and backends outlive their parents; only the proxy tier self-reaps


debugmcpdev:
## Summary

On Windows, every tier of the process tree except the innermost debug proxy outlives its parent. Observed after repeated `/testdebugger` runs: 45 `dev-proxy.mjs` supervisors accumulated (15 per Claude Code session — one per configured MCP variant), plus stdio-mode backends from 3-day-old sessions that nothing will ever reap. ~60 leftover node processes total.

## Mechanics (verified against source)

| Transition | Child survives? | Why |
|---|---|---|
| MCP client (claude.exe) dies → dev-proxy | **Survives** | `tools/dev-proxy/dev-proxy.mjs` has no stdin `end`/`close` handler, no `transport.onclose`, no exit hook — only SIGINT/SIGTERM handlers (`dev-proxy.mjs:774,780`), which a parent exit on Windows never delivers. `StdioServerTransport` + backend child keep the event loop alive forever. |
| dev-proxy dies → backend (http/sse) | **Survives** | Backend is spawned without an IPC channel (`stdio: ['ignore','pipe','pipe']`, `dev-proxy.mjs:171`) so it can't observe parent death; its `gracefulShutdown`/`closeAllSessions` is wired only to SIGINT/SIGTERM (`src/cli/http-command.ts:221-222`). |
| backend dies → debug proxies | **Cascades correctly** | Proxies are IPC children (`src/implementations/process-launcher-impl.ts:449`); parent death closes the channel and the proxy's `process.on('disconnect')` handler exits (`src/proxy/dap-proxy-core.ts:288-296`). The only tier with a working parent-death detector. |

Aggravating details:

- A new dev-proxy frees its port by netstat-killing whatever holds it (`_ensurePortFree`, `dev-proxy.mjs:500-549`) — so it reaps the *previous backend* but never the previous *supervisor*, and **stdio-mode backends (no port) are never reaped by anything**.
- dev-proxy kills backends with `taskkill /F` (`dev-proxy.mjs:450,468`), so the backend's graceful shutdown (and `closeAllSessions()`) never runs.
- `SessionManagerCore.handleTerminated/handleExited/handleExit` (`src/session/session-manager-core.ts:314-404`) null out `session.proxyManager` **without calling `.stop()`** — naturally-terminated sessions rely entirely on the proxy self-exiting.
[...]

PR #127, opened 2026-07-05
fix(session): stop proxy on natural session termination (part of #122)

debugmcpdev:
## Summary

PR 2 of 3 for #122. SessionManagerCore's `terminated`/`exited`/`error` handlers nulled out `session.proxyManager` without calling `stop()`, so naturally-terminated sessions relied entirely on the proxy worker self-exiting — and when a worker's shutdown stalls (hung adapter process), nothing reaps it. On Windows especially, orphaned proxies accumulate (~20 observed under one server generation).

## Change

The three handlers now fire-and-forget `proxyManager.stop().catch(log)` on the closure-captured instance after detaching listeners. `stop()` is idempotent against an already-exiting worker and force-kills after 5 s — belt and braces over the worker's own self-exit path.

Deliberately unchanged (and encoded in tests so it reads as intent, not omission):
- `handleExit` — the proxy process has already exited; `ProxyManager.handleProxyExit` already ran cleanup.
- `handleDryRunComplete` — the dry-run proxy exits itself with code 0.

## Testing

- Assertion-only additions across 4 session test files: `stop()` called exactly once on `terminated`/`exited`/`error`, and NOT called on raw `exit`.
- No existing expectations loosened — all 2298 pre-existing unit tests passed with the source change alone before test edits (stop() runs only after handlers detach, and its request-rejection matches what closeSession already did).
- Full unit suite: 147 files / 2298 tests passing; lint clean.

Part of #122: PR-1 (dev-proxy supervisor shutdown) landed in #125; PR-3 (backend orphan self-defense + graceful-then-force kill) follows.


debugmcpdev:
CI Container Tests caught a real regression: with real js-debug event ordering, 'terminated' races ahead of the final continue response, and the new stop() cancelled the in-flight request (and dropped its in-pipe response once isStopped was set). Fixed with a bounded drain-then-stop in ProxyManager.stop() (1s window; message processing stays live so in-pipe responses settle, typically within one 20ms tick), plus a post-drain null re-check for concurrent stops. Verified locally: docker-smoke-javascript 5/5 including both CI failures; full unit suite 2300 passing.

Issue #146, opened 2026-07-07
Proxy init failure dumps full environment (including secrets) into the MCP tool error response


debugmcpdev:
## Problem

When the proxy worker exits during initialization, the `Failed to attach: Proxy exited during initialization` error surfaced to the MCP client includes the proxy's whole stderr — and stderr contains the ProxyRunner debug line:

```
[DEBUG] [ProxyRunner] Raw message snapshot: { cmd: 'init', ... adapterCommand: { env: { GITHUB_PAT: 'github_pat_...', PATH: ..., 60+ vars } } }
```

i.e. the **entire adapter environment — including any secrets in the parent process env (e.g. `GITHUB_PAT`) — lands verbatim in the tool response** shown to the LLM client, and from there potentially into transcripts/logs.

## Repro

Any attach that dies during proxy init (e.g. the python attach deadlock, or attach to a dead port) returns the full env dump in the tool response.

## Suggested direction

- Don't log the raw init message env at all (log `envVars: <count>` like AdapterManager's spawn-config line already does), or redact values.
- Independently, consider truncating/sanitizing stderr before embedding it in user-facing error messages.

Found while E2E-verifying the #143 fix.

PR #162, opened 2026-07-09
test: add fast-check property-based tests (sanitizers, LineBuffer, DAP framing)

debugmcpdev:
## Summary
Adds `fast-check` and 12 property-based tests in 3 files — genuine coverage for the recently-hardened security-sensitive paths, and the integration OpenSSF Scorecard's Fuzzing check recognizes for JS/TS (currently 0/10, +0.51 aggregate expected):

- **`env-sanitizer.property.test.ts`** — generated secret shapes (`ghp_*`, `github_pat_*`, `sk-*`, `xox?-*`, `AKIA*`, PEM headers) never survive `sanitizeStderr`/`sanitizeStderrTail` wherever they sit on a line; env objects at any payload depth are always redacted by `sanitizePayloadForLogging`; `sanitizeEnvForLogging` preserves key sets and is idempotent. (Follows up the #146/#151/#153 leak audits.)
- **`line-buffer.property.test.ts`** — chunking invariance: any split of any input yields the same lines as the unsplit input (the exact guarantee the stderr redaction path depends on), plus the `maxPendingLength` memory bound with zero data loss.
- **`minimal-dap-framing.property.test.ts`** — generated DAP event streams survive arbitrary byte-level chunk splits (including mid-UTF-8-character) losslessly and in order.

## Verification
- All 12 properties pass; full unit suite: 154 files / 2491 tests green (~11s, no runtime impact worth noting).
- **Mutation check**: weakening a sanitizer regex (`AKIA[0-9A-Z]{16}` → `{99}`) fails the properties within ~10 generated cases; reverting restores green.


debugmcpdev:
The first CI run of these property tests **caught a real bug**: fast-check generated an env key named `__proto__`, which `sanitizeEnvForLogging` silently dropped (plain `obj[key] = value` routes through the prototype setter; in `redactEnvDeep` an object value would have replaced the output's prototype). Fixed in 9c64a8c by assigning via `Object.defineProperty`, with `__proto__`/`constructor`/`hasOwnProperty`/`toString` now permanently in the generator key pool plus an example-based regression test. Full unit suite green (154 files / 2492 tests).

PR #270, opened 2026-08-06
fix: JS breakpoint verified state + AX fixes from /testdebugger feedback (stop-reason normalization, Go locals, Java frames)

debugmcpdev:
## Summary

Fixes the JS breakpoint verified-state bug plus four agent-experience issues surfaced by two full `/testdebugger` runs (40/40 server×language combos each, including the extended attach/detach + `redefine_classes` workflows on the second run).

### JS breakpoint verified state (original branch commit)
- Make the child session authoritative for breakpoint verified state, so js-debug parent-session downgrades no longer flip verified breakpoints back to unverified.

### Stop-reason normalization (new `AdapterPolicy.normalizeStopReason` hook)
- New optional policy hook applied in the session-manager stopped handler before auto-continue gating, `lastStop` recording, and exceptionInfo enrichment. The pre-normalization reason is preserved as `lastStop.rawReason` when changed.
- **Rust/CodeLLDB**: an explicit pause is delivered via SIGSTOP and surfaced as `stopReason: "exception"`. Normalized to `"pause"` on SIGSTOP evidence (or an empty body while a pause is in flight). Real exceptions (SIGSEGV, panics) are never reclassified. Also stops the spurious `exceptionInfo` request on pause.
- **JavaScript/js-debug**: an explicit pause is emitted as `reason: "step"` with a body byte-identical to a genuine step, so normalization keys off the new `pausePending` session flag (set while a user-initiated pause request is in flight) instead of the body.
- `pause_execution` now reports `stopReason`/`rawStopReason` in its result (fresh stops only — stale stops are never echoed).

### Go: empty `get_local_variables` on optimized frames
- Delve names the locals scope `"Locals (warning: optimized function)"` for optimized frames; the policy's exact-equality match returned `[]` with no fallback. Now prefix-matched.
- `get_local_variables` reports the actual adapter scope name (warning suffix included) and adds a `warning` field with remediation guidance (`-gcflags="all=-N -l"` / debug mode).
- `examples/go/README.md` no longer shows `buildFlags` with `mode: "exec"` (Delve ignores it there).

### Java: pause landing in JDK-internal frames
- `isInternalFrame` matched `java.*` prefixes against frame names, but the JDI bridge emits bare method names (`sleep0`) — it could never match. It now also checks the declaring-type FQCN that lands in the `file` slot for no-debug-info frames (guarded by the existing FQCN heuristic), with an all-internal fallback so pure-JDK stacks still show frames. Locals now come from the first user frame automatically.
- Separable commit: the JDI bridge emits fully qualified frame names (`PauseTest.main`, `java.lang.Thread.sleep0`), matching JDWP/IDE conventions. Existing e2e assertions are substring-based and unaffected.

### Misc
- `get_local_variables` on a terminated session now explains the program finished and suggests `restart_debugging`, instead of the generic "must be paused at a breakpoint" message.
- The local `/testdebugger` command (gitignored `.claude/`, not in this diff) was extended to exercise `restart_debugging`, `list/remove/clear_breakpoints`, logpoints + `get_output`, `step_into`/`step_out`, `list_debug_sessions`, and opt-in attach/detach + `redefine_classes` workflows.

## Test plan

- [x] Full unit suite: 2796 passed (166 files); lint clean; `npm run build` clean
- [x] New unit coverage: Rust/JS `normalizeStopReason`, session-level normalization + `pausePending` lifecycle, `pause()` result fields, Go scope matching + real scope-name reporting + server `warning` field, Java internal-frame predicate + FQCN-in-file cases + all-internal fallback
- [x] Live verification via dev proxy: Go optimized exec-mode binary returns variables + warning; Rust pause → `{stopReason: "pause", rawStopReason: "exception"}`; JS pause → `{stopReason: "pause", rawStopReason: "step"}` with genuine steps still reporting `"step"`; Java pause → `PauseTest.main` with real locals
- [x] Two full `/testdebugger` runs (before and after the JS/terminated-message fixes): PASS across all 40 server×language combos, extended workflows 2/2 PASS, no regressions on the JS verified-state behavior this branch started with (`breakpointsReapplied: 1`, breakpoints re-hit after `restart_debugging` on all 6 variants)


debugmcpdev:
## Review — approve ✅

Reviewed all six commits (1,411-line diff) plus independent local verification on this branch: full unit suite (2,796 green), the three breakpoint-tooling e2e suites across all 8 languages (24/24 — the javascript legs exercise the child-authoritative verified-state path end-to-end), and the java smoke suite against the FQCN frame-name change (4/4). CI is green on both OSes including codecov/patch.

### What's strong

- **Child-authoritative verified state (js-debug)** is the right model and closes the loop #267 left open: child `setBreakpoints` responses become synthesized breakpoint events (all three mirror sites), so verification reaches the store without waiting for js-debug's async events; parent responses/events can upgrade but never downgrade, while id-matched downgrades (real child unbinding) still apply. The synthesized events also refresh stored `adapterId`s on every child re-send, which keeps subsequent events id-matchable — nice self-healing property.
- **`normalizeStopReason` hook** is well-placed (before auto-continue gating, `lastStop`, and exceptionInfo) and conservatively scoped: rust requires SIGSTOP evidence (or empty detail *while a pause is in flight*), js requires `pausePending`, so first-stop/entry handling and real exceptions can't be misclassified. Preserving `rawReason` keeps the response honest. Suppressing the spurious `exceptionInfo` request on normalized pauses is a good catch.
- **`pausePending` lifecycle** is airtight enough: set before the DAP request, cleared on every stop, on session-end, and on request failure. The pause-timeout path deliberately leaves it set — correct, since the adapter may still deliver that stop late, and whichever stop arrives next clears it.
- **Go/Java fixes** address root causes, not symptoms: `/^Locals?\b/` won't match `Localization` (test pins it); the FQCN-in-`file`-slot heuristic is guarded by the existing `isNonFileSourceIdentifier` and the all-internal fallback prevents empty stacks; reporting the adapter's *actual* scope name instead of the policy's canonical one is the honest contract.
- Test coverage is thorough and behavior-first at every layer, including the stale-stop guard (reference-comparison against `lastStopBefore`) and the non-mirroring-policy negative cases.

### Accepted edge cases (documented in code, agree with the trade-offs)

- Rust: a real exception with a completely empty description/text arriving while a pause is in flight would read as `pause`. CodeLLDB's SIGSEGV/panic stops carry descriptions, so this is theoretical.
- JS: a genuine step completing inside the pause window reports `pause` — as the comment says, that's what the user asked for.
- A stale child-event downgrade whose id no longer matches is ignored for mirroring policies; since every child re-send re-synthesizes verified events with fresh ids, staleness self-corrects.

### Completed in 2bc70e7c (pushed to this branch)

- CHANGELOG entries — the repo's convention for behavior-changing PRs (Added: stop-reason normalization + optimized-binary warning; Fixed: JS verified state, Java internal frames, terminated-session message).
- tool-reference: documented `data.stopReason` / `data.rawStopReason` on `pause_execution` and `lastStop.rawReason`.

Merging after CI re-runs on the docs commit.

Issue #265, opened 2026-08-05
adapter-rust: consolidate triplicated CodeLLDB platform-dir/candidate-path logic


debugmcpdev:
## Problem

The CodeLLDB platform-directory mapping and candidate-path-walking logic exists in **three independent copies** that can (and already did) drift:

1. **Async resolver** — `packages/adapter-rust/src/utils/codelldb-resolver.ts` (`resolveCodeLLDBExecutable()` / `getCodeLLDBVersion()`): platform/arch -> `platformDir` mapping + 4-candidate walk + `CODELLDB_PATH` fallback.
2. **Private sync copy** — `packages/adapter-rust/src/rust-debug-adapter.ts` (`resolveCodeLLDBExecutableSync()`, ~line 749): same mapping re-implemented with `existsSync`, a 3-candidate walk, and `this.platform` instead of `process.platform`.
3. **Vendor script** — `packages/adapter-rust/scripts/vendor-codelldb.js` `PLATFORMS` table: same five `platformDir` keys (`win32-x64`, `darwin-x64`, `darwin-arm64`, `linux-x64`, `linux-arm64`).

## Evidence it bites

The version fallback in `codelldb-resolver.ts` was hardcoded to `1.11.0` while the vendor script default had moved on to `1.11.8`. Fixed by exporting `DEFAULT_CODELLDB_VERSION` and adding a drift-guard test that greps the vendor script default (see `packages/adapter-rust/tests/codelldb-resolver.test.ts`), but the structural duplication remains.

## Subtlety for whoever picks this up

The sync copy's candidate list is NOT simply "missing a candidate" — it is depth-adjusted. `rust-debug-adapter.ts` compiles one directory shallower than `src/utils/codelldb-resolver.ts`, so its `'..'` candidate resolves to the same package-root `vendor/` directory that the async resolver reaches via `'..', '..'`. Any consolidation (e.g. a shared `buildCandidatePaths(baseDir, platform, arch)` helper) must take the caller's base directory as a parameter rather than using its own `__dirname`, or the two call sites will resolve different locations.

The `platformDir` mapping itself (platform/arch -> vendor key) has no such subtlety and is trivially shareable; the vendor script is plain JS run pre-build during postinstall, so it cannot import from `src/` TS — sharing with it likely means a JSON/`.cjs` constant or keeping the existing drift-guard-test approach.

## Suggested scope

[...]

Issue #260, opened 2026-08-04
Rust: panic pause reports lastStop.reason 'breakpoint' instead of 'exception'; CodeLLDB no longer advertises rust_panic


debugmcpdev:
Two related observations from the #244 validation sweep (2026-08-04, CodeLLDB 1.11.8):

1. **Reason label**: with the `rust_panic` filter armed (now the launch default), a `panic!` correctly pauses at the panic site with a full backtrace — but the stopped event carries `reason: 'breakpoint'`, not `'exception'`. CodeLLDB implements the panic filter as an internal breakpoint (on the `rust_panic` symbol) and labels the stop accordingly; we pass it through unnormalized. Callers following the documented contract (`lastStop.reason === 'exception'`) miss Rust panic pauses. Normalizing would need a policy-level stop-event hook, which `AdapterPolicy` doesn't have today (the same missing hook noted in #255's scoping) — worth designing once for both.

2. **Advertisement drift** (from the #243 drift warning, working as intended):
```
Exception filter drift: 'rust' policy declares [rust_panic] not advertised by the adapter (advertised: [cpp_throw, cpp_catch])
```
CodeLLDB 1.11.8 advertises only `cpp_throw`/`cpp_catch`, yet `rust_panic` demonstrably still functions (legacy acceptance). Keep the working filter, but the policy comment should note the discrepancy, and if a future CodeLLDB drops the legacy ID the drift warning is our early alarm.

Context: `lastStop.reason` mislabeling on Rust/Windows is a broader CodeLLDB theme — see the #255 trace where steps were labeled `breakpoint` and continue `step`.

Issue #236, opened 2026-08-03
[FEATURE] Breakpoint management: list_breakpoints / remove_breakpoint / clear_breakpoints tools


debugmcpdev:
## Motivation

Agents currently have write-only breakpoints: `set_breakpoint` exists, but there is no way to enumerate what is set, remove one, or clear all without tearing down the session. Long agentic debugging sessions accumulate stale breakpoints that cause confusing extra pauses; bisection workflows (move the window each iteration) need cheap removal.

## Proposed API

- `list_breakpoints {sessionId}` → `[{id, file, line, condition?, logMessage?, verified}]`
- `remove_breakpoint {sessionId, breakpointId}` (or `{file, line}`)
- `clear_breakpoints {sessionId, file?}` — all breakpoints, or all in one file

## Implementation sketch

Breakpoints are already tracked per-session in the session store (queued + verified state). Listing is a read of that store; removal re-sends DAP `setBreakpoints` for the affected file with the remaining set (DAP replaces the file's whole array, so removal = re-send minus one).

## Acceptance criteria

- list reflects verified state and includes adapter-assigned ids.
- remove/clear take effect immediately while paused or running.
- Unit tests over the session store diff logic; e2e for at least Python + mock adapter.

PR #277, opened 2026-08-06
fix(#255): correct the continue-re-stop diagnosis — multi-location macro breakpoints

### LINKED ISSUE #255: Rust/Windows: continue_execution re-stops at the same breakpoint (never advances without a step_over)
## Context

Found while live-verifying PR #254 on Windows (CodeLLDB 1.11.8, MSVC-built example with `LLDB_USE_NATIVE_PDB_READER=1`).

## Problem

With a breakpoint at `examples/rust/hello_world/src/main.rs:26`, once the session is paused at that breakpoint, `continue_execution` succeeds (`success: true`) but the session immediately re-stops at the **same** breakpoint (same line, same threadId, `stopReason: "breakpoint"`, fresh `lastStop` timestamp). Repeats indefinitely — the program never advances past the line by continue alone.

Reproduced twice in one session:
- 19:54:26 paused at main.rs:26 (breakpoint)
- `continue_execution` → 19:55:19 paused again at main.rs:26 (breakpoint)

## Workaround

`step_over` once (moves to line 27), then `continue_execution` runs to completion normally.

## Notes

- Likely LLDB-on-Windows failing to step over the re-armed breakpoint trap before resuming (classic breakpoint-re-hit behavior), possibly specific to the native PDB reader path.
- Not covered by existing e2e: `tests/e2e/mcp-server-smoke-rust.test.ts` inspects locals at the breakpoint but never continues past a user breakpoint to completion; the async test only steps.
- The new Rust `get_output` e2e assertion in PR #254 was deliberately written to not require continuing past a breakpoint because of this quirk.
- Worth checking whether the first "paused, reason: breakpoint" reported by `start_debugging` is actually the Windows initial system stop being mislabeled — in the same session, the real user breakpoint was only reached after one continue (see skill note "Windows initial stop" in skills/debugging/references/rust.md:55).

debugmcpdev:
## Summary

Validating the Rust trio (#272/#273/#274/#276) on Ubuntu surfaced a failing test on `main`: the strict non-win32 reproducer added in #274 asserts that **one** `continue_execution` from a breakpoint at `examples/rust/hello_world/src/main.rs:26` runs the program to completion. It does not, on Linux — but not because of the bug #255 describes.

## Root cause

Line 26 is `let message = format!("Language: {}, Version: {}", name, version);`. The `format!` expansion inlines several call sites onto that one source line, so LLDB plants a breakpoint **location** at each — the `setBreakpoints` response says `Resolved locations: 3`. Each continue advances to the next location, so the session re-pauses on the same `file:line`, with the same breakpoint id, at a **different program counter**. The line takes three continues to leave.

Confirmed with a raw DAP client driving the vendored CodeLLDB directly (mcp-debugger entirely out of the loop):

| breakpoint line | source | resolved locations | continues to completion |
|---|---|---|---|
| 26 | `let message = format!(...)` | 3 | 3 |
| 13 | `let name = "Rust";` | 1 | 1 |
| 42 | `let sum = a + b;` | 1 | 1 |

Each intermediate stop reports `reason: "breakpoint"`, `hitBreakpointIds: [1]`, `line: 26`, `col: 19` — at PCs `…C6FF`, `…C73A`, then termination. Also ruled out the #236 post-stop `setBreakpoints` re-sync by disabling it: no change.

So this is ordinary LLDB multi-location behavior on **every** platform — not Windows-specific, not MSVC/PDB-specific, and not ours. The prior "Linux remains clean" note only held because nothing in the suite had ever continued past a user breakpoint on Linux, which is the exact gap #274 set out to close.

## Changes

- **`tests/e2e/mcp-server-smoke-rust.test.ts`** — replaces the wrong-premise reproducer with two tests sharing one launch helper, both running on all platforms (the win32 special-casing and its `step_over` workaround are gone):
  - a **single-location** line (42) that must complete in exactly one continue — the coverage gap that let this survive;
  - a **multi-location** macro line (26) that must stay on line 26 between stops, drain its locations within a bound, and then exit 0 in more than one continue.
- **`skills/debugging/references/rust.md`** — the quirk bullet and troubleshooting row described a Windows-only re-hit bug with a `step_over` workaround; they now describe multi-location macro breakpoints and the correct guidance (keep continuing, or `step_over` once to cross the whole line).
- **`mcp-server-breakpoint-management.test.ts` / `mcp-server-smoke-restart.test.ts`** — comments corrected. Behavior deliberately untouched: both already tolerate multi-location stops (one clears breakpoints first, the other loops), and Windows can't be exercised from this machine.

## Testing (Ubuntu 24.04, x86_64, CodeLLDB 1.11.8, rustc 1.91.1)

- `mcp-server-smoke-rust` 4/4, `mcp-server-breakpoint-management` 8/8, `mcp-server-smoke-restart` 8/8
- Full unit + integration with `LEAK_GUARD_STRICT=1`: 2852 passed
- Lint and typecheck clean

## Also verified on Linux while here (no changes needed)

- **#260** — the Linux leg PR #273 deferred to "ubuntu CI", which never runs it: `test:ci-coverage` is `--project unit --project integration`, so **no e2e runs in CI at all**. The panic trio passes; Linux matches the Windows capture exactly — stopped body `{"allThreadsStopped":true,"hitBreakpointIds":[1],"reason":"breakpoint","threadId":…}` with no description/text, session reporting `{reason: 'exception', rawReason: 'breakpoint'}`. An ordinary user breakpoint still reports `breakpoint`, so the id-disjointness discriminator does not false-positive here.
- **#265** — real Rust sessions launch, so the consolidated resolver genuinely finds the vendored `linux-x64` CodeLLDB (the new unit tests are hermetic and can't show this).
- **#276** — POSIX pause unaffected: `stopReason: 'pause'`, `rawStopReason: 'exception'`.
- **#274**'s cross-language `'unknown'` fallback — 42 e2e tests across python/javascript/go/java/ruby/dotnet/mock/restart/logpoints/breakpoint-management, all green.

Closes #255


debugmcpdev:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Verified on the platform this PR could not exercise: all 4 tests in [tests/e2e/mcp-server-smoke-rust.test.ts](https://github.com/debugmcp/mcp-debugger/blob/1134743a0d71a291aae435f41b7c5db83e61c649/tests/e2e/mcp-server-smoke-rust.test.ts) pass on Windows 11 (GNU toolchain, CodeLLDB 1.11.8) — the single-location line-42 test completes in exactly one continue, and the line-26 macro test drains its locations and exits 0. This confirms the diagnosis on Windows: the re-stop follows breakpoint-location count, not the OS.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #243, opened 2026-08-03
[FEATURE] Capture adapter initialize capabilities + best-effort exceptionInfo enrichment on exception stops


debugmcpdev:
Follow-up to #220 (breakOnExceptions shipped without runtime capability negotiation).

## Gaps

1. **The DAP `initialize` response is discarded** (`dap-proxy-connection-manager.ts`, `initializeSession`) — the adapter's real `Capabilities` (`exceptionBreakpointFilters`, `supportsExceptionInfoRequest`, `supportsExceptionFilterOptions`, ...) are unavailable at runtime. #220's per-language filter tables live on the adapter policies instead; capturing capabilities would let us validate the resolved filters against what the live adapter actually declares (and warn on drift — the static `getCapabilities()` declarations have already drifted once, see the js-debug correction in #220's PR).

2. **DAP `exceptionInfo` is never requested.** On a `stopped` with `reason: 'exception'`, a best-effort `exceptionInfo` call (gated on the captured `supportsExceptionInfoRequest`, failures swallowed) could enrich `lastStop` with `exceptionId`, `breakMode`, and the full stack description beyond the `description`/`text` fields the stopped event carries today. debugpy, js-debug, and netcoredbg all support it.

## Sketch

- Capture the initialize response body in the worker, forward it to the SessionManager in the `adapter_configured` status (or a dedicated message), store on `ManagedSession`.
- On exception stops, worker (or SessionManager via `sendDapRequest`) issues `exceptionInfo(threadId)` when supported and merges the result into `lastStop`.

PR #301, opened 2026-08-11
feat: DAP mirror — read-only IDE attach to agent-owned debug sessions (#217)

### LINKED ISSUE #217: [FEATURE] DAP mirror: let an IDE attach to an agent-owned debug session with shared pause state
### The workflow this enables

An agent debugs a process that **no IDE owns** — in CI, in a container, on a remote host, or driven from a plain terminal — parks it at an interesting point, and a human then attaches their IDE (VS Code or any DAP client) to inspect the **live, paused** state: walk the actual stack, evaluate expressions, examine real objects. Strictly better than a core dump or logs.

Originally requested in #7 ("connect VS Code to a Python session started from MCP"). For users who already have VS Code open, that literal ask is now best served the other way around — run the session in VS Code and let the agent drive it via VS Code's agent-debugging/MCP surface. This issue tracks the scenario that inverted flow structurally *cannot* cover: sessions that originate where no IDE exists.

### Motivating scenarios

- **CI flake triage**: the agent reproduces a flaky test under the debugger, catches the failure live, and holds the process paused; a human attaches remotely in the morning and inspects the actual failing state.
- **Long-running process babysitting**: an agent attached to a staging service overnight notices an anomaly, pauses at it, and hands a human the live session.
- **Terminal / chat-driven users**: someone debugging via an agent with no editor session at all occasionally wants to drop into an IDE for a look around, without restarting the run.

### Why this doesn't work today

1. Each launch-mode session spawns its adapter on a **private ephemeral localhost port** (e.g. `python -m debugpy.adapter --host 127.0.0.1 --port <ephemeral>`) with mcp-debugger as the sole DAP client — there is no socket an IDE could reach.
2. Debug adapters are generally **single-client** (debugpy explicitly so); a second simultaneous DAP client isn't something we can configure around.
3. Sequential hand-off exists in attach mode (agent attaches to a `debugpy --listen` process, detaches, human attaches) but **the pause doesn't survive the swap** — on disconnect the debuggee resumes, which loses the "parked at the interesting point" state that is the whole value.

### Proposed direction: a DAP mirror

mcp-debugger already sits mid-stream as a DAP client. It could additionally expose a **DAP server endpoint** per session that an IDE connects to, multiplexing both clients onto the single live adapter session:

- Forward `stopped`/`continued`/`thread`/`output` events to both clients; the paused frame is genuinely shared.
- Reconcile request IDs and route responses to the requesting client.
- Mediate breakpoint ownership (union of both clients' breakpoints, or agent-owned with IDE additions).
- Per-client capability negotiation (the IDE's `initialize` is answered from the session's known capabilities).

### Open questions

- Read-only mirror first (inspection only: stack/scopes/variables/evaluate) vs. full dual-control from day one — read-only sidesteps most of the ownership/reconciliation complexity and covers the motivating scenarios.
- How the endpoint is exposed and secured (DAP sockets are unauthenticated; localhost + user-initiated `expose` tool call, with the same port-forward guidance as remote attach).
- Per-adapter quirks (js-debug child sessions, Java JDI bridge).

## 📊 Impact

- **Who would benefit**: anyone combining agent-driven debugging with human inspection outside the IDE-owned flow
- **Priority**: Medium
- **Complexity**: Large

debugmcpdev:
Implements the v1 (read-only) DAP mirror from the design spike in #217: an agent debugs a process no IDE owns — CI, a container, a plain terminal — parks it at an interesting point, and a human attaches VS Code (or any DAP client) to the **live, paused** session and inspects real state. Execution control stays with the MCP session.

## What's in the box

**Two new MCP tools (26 → 28)**
- `expose_session {sessionId}` → `{host: "127.0.0.1", port, token}` — starts a loopback-only, token-gated DAP server endpoint for the session; idempotent (same endpoint + token on re-call). The result `message` carries a ready-to-paste VS Code config (`"debugServer": <port>` + `"mirrorToken"`), with a container-networking note appended in container mode.
- `unexpose_session {sessionId}` — closes the endpoint and disconnects IDE clients (`terminated` event); no-op success when not exposed.

**Worker-hosted mirror** (`src/proxy/dap-mirror-server.ts`) — lives in the proxy worker, the sole owner of the live adapter connection (one worker = one session = one mirror):
- `initialize` answered from the adapter's cached capabilities with control affordances masked off (restart/step-back/setVariable/etc. forced false, `exceptionBreakpointFilters: []`), so IDE UIs degrade *before* the user can click.
- Token checked timing-safely at `attach`/`launch` (`crypto.randomBytes(24)` base64url); bad token → user-visible error + socket close; 30s auth timeout; max 4 clients.
- **Forwarded** (via the worker's `IDapClient.sendRequest`, so js-debug child routing comes free): `threads`, `stackTrace`, `scopes`, `variables`, `source`, `evaluate`, `exceptionInfo`, `loadedSources`, `modules`.
- **Soft-succeeded** so VS Code's attach handshake survives: `set*Breakpoints` (reported `verified:false` with an explanatory message), `setExceptionBreakpoints`.
- **Rejected** (default-deny, quiet `showUser:false` errors): `continue`, `next`, `stepIn/Out`, `pause`, `setVariable`, `restart`, `terminate`, and everything unlisted.
- Per-client seq spaces with correct `request_seq` correlation under out-of-order completion.
- **Late join**: on attach while paused, the mirror synthesizes a `stopped` event from the worker's recorded last stop, so the IDE lands directly on the paused frame (with a 1s fallback for clients that skip `configurationDone`).

**Event plumbing** — `stopped`/`continued` fan out from the worker's own handlers (after the threadId backfill for Delve/JDI; the generic `'event'` channel would race the async backfill and ship a threadId-less body), everything else rides `MinimalDapClient`'s generic `'event'` feed. A successful `continue`/step *response* also synthesizes `continued` for mirror clients when no event arrives — DAP clients are entitled to infer resumption from the response, and stale paused frames in the IDE were the alternative.

**Transport: zero new IPC surface** — `mirrorExpose`/`mirrorUnexpose` ride the existing correlated `dap` envelope (precedent: `redefineClasses`), intercepted at the top of `handleDapCommand` before the connectivity bail and policy queueing.

**Shared framing codec** — `MinimalDapClient`'s battle-tested Content-Length decoder is extracted verbatim into `src/proxy/dap-framing.ts` (`DapFrameDecoder` + `encodeDapMessage`) and reused by client and mirror; the existing framing property suite now runs against the extraction as a regression net, plus a new property suite drives the decoder directly.

**Security**
- Bind hard-coded to `127.0.0.1`; token required on every client.
- The shared payload sanitizer now redacts string `token`/`mirrorToken` values, and ProxyManager's raw received-message debug log is sanitized (it previously dumped every IPC message verbatim — the `dapResponse` carrying the token would have hit debug logs).
- `SECURITY.md` + tool-reference call out that the token is a **debuggee-execution capability** (evaluate is forwarded), not a view-only credential.
- `list_debug_sessions` shows `exposure: {host, port}` — never the token — gated on a running proxy so stale records can't surface after teardown paths that skip cleanup.

## Testing

- `tests/proxy/dap-mirror-server.test.ts` — 46 hermetic cases (fake sockets feeding the real decoder): handshake + capability mask, token gate, auth timeout, full reject-table sweep, allowlist forwarding with seq correlation, out-of-order completions, two-client fan-out with independent seq spaces, late-join synthesis + fallback timer, terminated dedupe, max clients, malformed frames.
- `tests/proxy/dap-proxy-worker.test.ts` — 14 new worker cases: interception before queueing, idempotent expose, host reflects live worker state, resume inference, shutdown notification, capability retention through the #243 once-only guard.
- `tests/core/unit/{server,session}` — tool dispatch/result shaping (incl. container note) and exposeSession/unexposeSession logic (stale-record cleanup, relaunch reset, token absent from projections).
- `tests/integration/dap-mirror/` — live-socket check of the production `net` wiring.
- `tests/e2e/mcp-server-dap-mirror.test.ts` — mock-adapter full stack with a scripted TCP DAP client: full inspection surface, control rejection leaves the session paused, token rejection, two concurrent clients, idempotency, endpoint closed after unexpose/close.
- Full suite: **3300 passed, 0 failed**; lint + strict tsc clean.
- **Live dogfood** against a real debugpy session via the dev proxy: exposed a paused Python session, attached a scripted IDE client with the token — masked capabilities, late-join `stopped`, real locals, `evaluate` forwarded, `continue` rejected with the session still agent-controlled, unexpose → `ECONNREFUSED`, and the token appears in **zero** log files (proxy log, DAP trace, debugpy logs).

## Out of scope (per the spike's phasing)

v2 breakpoint union and opt-in execution control; per-language polish (js-debug child-session guarantees, JDI bridge notes) — the mirror routes through `sendRequest`, so js-debug child routing already works, but v1 makes no per-adapter guarantees beyond the tested mock + Python paths.

Closes #217


debugmcpdev:
Added a coverage pass before merge (commit c0687140) closing four gaps a review of the suites turned up:

1. **Real handler wiring** — the worker's `stopped`/`continued` mirror fan-out is now tested through the actual `setupDapEventHandlers` closures (incl. the Delve/JDI threadId-backfill path), replacing a simulation that would have survived deletion of the production code.
2. **Live mirroring e2e** — an attached IDE client now observes MCP-side stepping (synthesized `continued` + fresh `stopped(step)`) and debuggee exit (exactly one `terminated` through the terminal-signal dedupe, then socket close + port refused). This is the shared-pause-state promise itself, previously only tested against a static pause.
3. **restart_debugging lifecycle** — e2e now proves a restart kills the old endpoint, clears the `exposure` projection, and a fresh expose binds a new port.
4. **Token-redaction regression test** — ProxyManager's received-message debug log is asserted to redact the mirror token (the leak this PR fixes).

Full suite green locally via the pre-push hook; CI re-running.

Issue #318, opened 2026-08-12
[BUG] ruby launch ignores launchConfig.env — user-provided env never reaches the debuggee


debugmcpdev:
Found while investigating #317.

**What happens**: on the ruby launch path, `env` passed via `dapLaunchArgs` is silently dropped. `RubyDebugAdapter.buildAdapterCommand` builds the spawn environment solely from `{...process.env, RUBY_DEBUG_DAP_SHOW_PROTOCOL}` and never reads `config.launchConfig.env`. The value does survive `transformLaunchConfig` into the DAP launch request args, but that's a dead end: with `rdbg --open -c -- ruby <script>` the debuggee process is already running by the time the DAP `launch` request arrives, so rdbg's launch response is effectively an ack and the env is never applied.

**Expected**: `launchConfig.env` merged into the spawn env of the rdbg child (e.g. `{...process.env, ...launchConfig.env, <adapter-managed vars>}`), so users can set env for the debuggee the same way as other adapters. Merge order matters — an explicit user value should win over inherited `process.env`, and adapter-managed vars should be composed rather than clobbered.

**Contrast**: attach mode *does* forward env (`if (config.env) attachConfig.env = config.env` in `transformAttachConfig`), and other adapters pass user env through their spawn paths.

**Where**: `packages/adapter-ruby/src/ruby-debug-adapter.ts` — `buildAdapterCommand` (env literal) and `buildTargetCommand`.

Environment: main, all platforms (structural, not platform-specific).

PR #319, opened 2026-08-12
fix(ruby): stream debuggee stdout mid-run via injected sync prelude (#317)

### LINKED ISSUE #317: [BUG] comprehensive e2e matrix: ruby get_output marker '1: 1' not captured mid-run (soft FAIL cell)
The comprehensive e2e matrix (tests/e2e/comprehensive-mcp-tools.test.ts) consistently reports a soft FAIL for ruby get_output on Windows:

```
[✗] get_output (ruby): marker "1: 1" not captured (entries=1)
```

The cell is reproducible across runs (with and without coverage instrumentation) but is invisible in CI/vitest because the matrix records it softly — the vitest test itself passes, so nothing goes red.

**What happens**: in the workflow sequence (breakpoint in the fizzbuzz loop → steps/evaluate → continue → 2s wait → get_output), only one output entry (the rdbg REPL banner) has arrived by the time get_output runs. Iteration 1's `puts` output never arrives as a DAP output event mid-run.

**What works**: launching the same fizzbuzz.rb without the workflow and letting it run to termination captures everything — banner, `fizzbuzz complete` (stderr), and all 15 stdout lines, with the stdout entries timestamped together at exit (the #254/#258 exit-flush drain). So capture and forwarding are intact; the gap is specifically *mid-run* delivery while the session sits paused in the loop — consistent with Ruby block-buffering stdout on a pipe (~8KB buffer that a few fizzbuzz lines never fill), with the flush only happening at exit.

**Why it matters**: the outputMarker was added in PR #254 with the comment "iteration 1's puts — the loop breakpoint re-arms, so later output isn't guaranteed", i.e. iteration 1 *was* expected to be captured mid-run. Either that expectation regressed at some point, or it was only ever true on Linux/CI. Unknown since when — the soft cell can't fail a build, so nobody was looking.

**Possible directions**:
- Force `STDOUT.sync = true` (or `$stdout.sync`) in the rdbg launch path so debuggee output streams instead of buffering, or
- Relax the matrix expectation for ruby to "captured after termination" and make the cell hard-assert that, or
- Make the matrix cell hard-fail so regressions in it are visible.

Ruled out: the #237 redaction hook (session-manager-core handleOutput) — it transforms events that arrive and cannot drop them; the missing marker means the event never arrived, and the arrival path (adapter stdio forwarding, proxy) is untouched by #237. The same hook captured all 17 entries in the direct repro.

Environment: Windows 11, Ruby 3.4 + rdbg, main @ 8c5c2db7.

debugmcpdev:
Fixes #317.

## Problem

Ruby block-buffers `$stdout` when it is a pipe (`$stdout.sync == false`; `$stderr` defaults to `sync == true` — which is exactly why stderr lines arrived mid-run while stdout didn't). `rdbg -c` hands the debuggee the adapter process's piped stdio, so `puts` output only reached the proxy's stdio scraper at process exit (the #254/#258 exit-flush drain) — never while the session sat paused at a breakpoint. The comprehensive matrix's ruby `get_output` cell soft-failed on exactly this: the workflow pauses at a re-armed loop breakpoint, the process never exits, nothing flushes (`entries=1`).

## Fix

`buildTargetCommand` now materializes a two-line prelude (`$stdout.sync = true; $stderr.sync = true`) in the session log dir and injects it into the target ruby argv as a single `-r<path>` element, in both the plain and bundler branches.

Design notes:
- **Argv insertion, not RUBYOPT**: a single argv element is space-safe end-to-end (spawn argv array; rdbg `-c` execs the command after `--` verbatim). `RUBYOPT` splits on whitespace and would break for install paths containing spaces.
- **Session log dir, not `os.tmpdir()`**: the file is `require`d into the debuggee; a predictable name in shared world-writable `/tmp` would be a local code-injection surface on POSIX. The log dir is product-owned. Runtime generation also means no npm-bundle asset plumbing.
- **Graceful degradation**: if the prelude can't be written, the launch proceeds without it (exit-only flushing, as before) and logs an error.
- **Unconditional, launch mode only** — same spirit as Python's `PYTHONUNBUFFERED=1`. A script can set `$stdout.sync = false` itself to opt out. Attach connects to a process the server did not start, so nothing is injected there (documented).
- Verified empirically that rdbg's stop-at-load still lands on the main script (line 1), not the prelude.

## Tests

- **New hard e2e regression test** (`mcp-server-smoke-ruby.test.ts`): breakpoint in the fizzbuzz loop → continue → poll `get_output` for iteration 1's marker → assert it arrives **while the session is still paused**, so the exit-flush path cannot satisfy it. Red against the pre-fix build (poll timeout), green after (<2s).
- Unit tests for prelude materialization (idempotent reuse, tampered-content rewrite, intermediate dirs, graceful `null` on write failure) and argv shape (plain, bundler, degraded).
- Integration args-tail assertion updated (`ruby-session-smoke.test.ts` — runs in CI with the real `buildAdapterCommand`).
- Comprehensive matrix: ruby `get_output` flips FAIL→PASS — **175 PASS / 0 FAIL / 6 SKIP** (was 174/1/6); the matrix cell stays soft by design, the new e2e test is the hard guard.
- Full pre-push suite green on Windows; docs updated (`docs/ruby/README.md` Program output).

Follow-up filed: #318 (`launchConfig.env` silently ignored on the ruby launch path — pre-existing, out of scope here).

### DIFF
--- a/docs/ruby/README.md
+++ b/docs/ruby/README.md
@@ -74,6 +74,15 @@ events. The proxy forwards those lines as synthesized `stdout`/`stderr` entries,
 output as usual; rdbg's own `DEBUGGER:` stderr banners are excluded and only appear in
 the session log.
 
+Because Ruby block-buffers `$stdout` when it is a pipe, launch mode also injects a small
+prelude (via `ruby -r`) that sets `$stdout.sync = true` and `$stderr.sync = true` before
+the script runs. Without it, `puts` output would only appear in `get_output` after the
+process exits; with it, output streams in near-real-time — including while the session is
+paused at a breakpoint (issue #317). A script that genuinely needs buffered stdout can
+set `$stdout.sync = false` itself. Attach mode connects to a process the server did not
+start, so no prelude is injected there — set `$stdout.sync = true` in your program if you
+need mid-run output while attached.
+
 ### Bundler projects
 
 Pass `useBundler` through the launch configuration to run the target via `bundle exec`:
--- a/packages/adapter-ruby/src/ruby-debug-adapter.ts
+++ b/packages/adapter-ruby/src/ruby-debug-adapter.ts
@@ -28,7 +28,8 @@ import {
   getRubyVersion,
   getRdbgVersion,
   getRubySearchPaths,
-  buildRdbgInvocation
+  buildRdbgInvocation,
+  ensureRubySyncHelper
 } from './utils/ruby-utils.js';
 
 interface RubyPathCacheEntry {
@@ -267,11 +268,22 @@ export class RubyDebugAdapter extends EventEmitter implements IDebugAdapter {
   }
 
   private buildTargetCommand(config: AdapterConfig, launchConfig: RubyLaunchConfig): string[] {
+    // Ruby block-buffers $stdout when it is a pipe, and rdbg -c hands the
+    // debuggee the adapter process's piped stdio — without intervention,
+    // puts output only reaches the proxy's stdio scraper at process exit
+    // (issue #317). Inject a tiny -r prelude enabling sync mode. A single
+    // argv element is space-safe (spawn argv array; rdbg execs the command
+    // after `--` verbatim), unlike RUBYOPT which splits on whitespace. On
+    // failure, launch proceeds with exit-only flushing.
+    const syncHelper = ensureRubySyncHelper(config.logDir, this.dependencies.logger);
+    const rubyArgs = syncHelper ? [`-r${syncHelper}`] : [];
+
     if (launchConfig.useBundler) {
       return [
         launchConfig.bundlePath || 'bundle',
         'exec',
         config.executablePath,
+        ...rubyArgs,
         config.scriptPath,
         ...(config.scriptArgs || [])
       ];
@@ -279,6 +291,7 @@ export class RubyDebugAdapter extends EventEmitter implements IDebugAdapter {
 
     return [
       config.executablePath,
+      ...rubyArgs,
       config.scriptPath,
       ...(config.scriptArgs || [])
     ];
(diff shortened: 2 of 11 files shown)


debugmcpdev:
Post-PR multi-angle self-review outcome (all angles verified against the real Ruby 3.4/debug-1.11 toolchain):

**Applied** (second commit):
- `pollUntil` was a third verbatim copy across e2e suites — promoted to `tests/e2e/smoke-test-utils.ts`, all three suites now import it.
- The ruby integration smoke test's `logDir` pointed at `<cwd>/logs/tests`, which `buildAdapterCommand` now genuinely writes into — switched to a per-test temp dir (hermetic, and the `-r` assertion no longer depends on repo-tree writability).

**Dismissed with rationale**:
- *Dry-run writes the helper file*: `dryRunSpawn` already creates the run log dir today; materializing the command is the point of a dry run, and the 40-byte write is part of the command. Benign.
- *`mkdirSync` redundant under SessionManager*: deliberate — it keeps `ensureRubySyncHelper` self-contained for direct callers (unit/integration tests, future adapters).
- *Read-compare-rewrite "dead in production"*: per-run dirs make it rare, but shared-dir callers (tests, custom logDir configs) do exercise it, and it keeps concurrent same-content writes benign.
- *Hardcoded `line: 15` vs `RUBY_BP_LINE`*: matches the existing test in the same file; importing `language-matrix-utils` would execute module-level toolchain probes as an import side effect.
- *Helper lives under the session log dir whose default root is `os.tmpdir()`*: acknowledged nuance to the "product-owned dir" rationale on multi-user Linux — an attacker owning the first-come `/tmp/debug-mcp-server` parent could race the write. Reviewers judged the runtime-write approach defensible (Ruby has no `PYTHONUNBUFFERED` equivalent; static assets don't survive the esbuild-only npx bundle without new plumbing). A dedicated assets dir / shipped static asset is the future hardening if we ever want it.

PR #408, opened 2026-08-22
perf(startup): lazy-import SSE/HTTP transport stacks in CLI actions

### LINKED ISSUE #400: perf(startup): stdio mode eagerly imports express and all three transport stacks
## Problem

`src/index.ts:41-60` statically imports all three transport command modules, so a stdio-mode server (the overwhelmingly common case — every `claude mcp add` config) loads the entire HTTP stack it will never use:

- `./cli/sse-command.js:2-3` → **express** + `@modelcontextprotocol/sdk/server/sse.js`
- `./cli/http-command.js:3-7` → **express**, `sdk/server/streamableHttp.js`, `sdk/server/express.js`
- both are imported eagerly regardless of which subcommand runs

Express and its dependency tree (body-parser, qs, send, etc.) are parsed and instantiated into the module graph on every stdio startup — pure baseline RSS with zero benefit.

## Proposal

Lazy-import the transport modules inside their commander action handlers:

```ts
.action(async (options, command) => {
  const { handleHttpCommand } = await import('./http-command.js');
  ...
})
```

`check-rust-binary` already demonstrates the pattern in this codebase (`src/cli/commands/check-rust-binary.ts:3,8-25` — type-only import at module scope, real `import()` deferred into the handler).

Notes:

- The Docker bundle (`dist/bundle.cjs`) is a single esbuild artifact; lazy `import()` of code inside the same bundle still avoids *executing* the express module graph even though the bytes ship. The npx tsup bundle behaves the same. So the win is parse/execute + heap, not download size.
- Measure with `scripts/mem-bench.mjs --scenario idle` (dist and bundle targets) before/after.

Sibling of the adapter-loader probe issue (first `tools/list` imports all 9 adapter packages) — together these two bound the "server that has answered `initialize` and one `tools/list`" footprint.

debugmcpdev:
Fixes #400

stdio mode — every `claude mcp add` config — statically imported all three transport command modules, dragging express and the SDK's `streamableHttp`/`sse`/`express` stacks (body-parser, qs, send, …) into the module graph of a server that never uses them.

## Change

- `src/index.ts`: the SSE and HTTP command modules are now imported with `await import()` inside their commander action closures, so they only load when that subcommand actually runs (same pattern as `check-rust-binary`). The stdio command module stays static — it has no HTTP dependencies and is the default path.
- The `handleSSECommand`/`handleHttpCommand` re-exports from `index.ts` are dropped. No consumers existed: tests and tooling import the command modules directly.
- `src/cli/stdio-command.ts`: `StdioOptions` becomes an `import type` (it's a type-only use).

## Laziness is locked in by a test

`tests/unit/index-lazy-imports.test.ts` mocks `express` with a factory that flips a flag on evaluation, imports the entry point, and asserts the flag stayed false. Before this change the test fails (express evaluates on import); after, it passes. `vi.mock` intercepts dynamic imports too, so the existing `index.test.ts` handler-closure assertions keep working unchanged.

## Measured (mem-bench, this machine, Windows)

`node scripts/mem-bench.mjs --target dist --scenario idle` — median RSS at `after-initialize`, 5 trials:

| | RSS med | peak med |
|---|---|---|
| main (`2014358c`) | 108.5 MB | 108.5 MB |
| this branch | **97.4 MB** | 97.5 MB |

**−11.1 MB (−10.2%) idle stdio footprint.**

Note: in the docker/npx single-file bundles the bytes still ship, but the express module graph is no longer parsed/executed on stdio startup — the win is parse/execute + heap, as scoped in the issue.

## Verification

- Full `npm test` green (259 files / 3947 tests), `npm run lint` clean
- `tests/unit/index.test.ts` + `tests/unit/cli/*` pass unchanged

### DIFF
--- a/src/cli/stdio-command.ts
+++ b/src/cli/stdio-command.ts
@@ -1,7 +1,7 @@
 import type { Logger as WinstonLoggerType } from 'winston';
 import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
 import { DebugMcpServer } from '../server.js';
-import { StdioOptions } from './setup.js';
+import type { StdioOptions } from './setup.js';
 import type { ProcessLike } from '../interfaces/process-interfaces.js';
 
 export interface ServerFactoryOptions {
--- a/src/index.ts
+++ b/src/index.ts
@@ -51,8 +51,6 @@ import {
   setupCheckRustBinaryCommand,
 } from './cli/setup.js';
 import { handleStdioCommand } from './cli/stdio-command.js';
-import { handleSSECommand } from './cli/sse-command.js';
-import { handleHttpCommand } from './cli/http-command.js';
 import { handleCheckRustBinaryCommand } from './cli/commands/check-rust-binary.js';
 import { getVersion } from './cli/version.js';
 import fs from 'fs';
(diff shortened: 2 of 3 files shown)


debugmcpdev:
Post-merge measurement note: the −11.1 MB RSS delta above was measured while full test suites ran concurrently, and Windows had trimmed the bench processes' working sets — RSS understated steady-state. Re-measured on a quiet machine (main @ f037e994): idle stdio sits at ~108 MB RSS ≈ private bytes. The durable win from this PR is the **private-bytes** delta visible in the same runs: 111.7 → 106.7 MB (~−5 MB, the express/transport module graph). The laziness itself is locked in by `index-lazy-imports.test.ts` either way.

PR #473, opened 2026-08-24
Fix: Clear provisional js-debug breakpoint message when verified

### LINKED ISSUE #471: Verified JS breakpoints keep the raw provisional message 'breakpoint.provisionalBreakpoint' (untranslated key + contradicts verified:true)
## Summary

A JavaScript breakpoint that has successfully bound still carries the *provisional* (pre-bind) message from js-debug, and that message is an untranslated l10n key. `list_breakpoints` ends up asserting two contradictory things at once:

```jsonc
{
  "id": "0bbdc056-…",
  "file": ".../examples/javascript/simple_test.js",
  "line": 11,
  "logMessage": "LOGPOINT a={a}",
  "verified": true,                              // ← bound
  "adapterId": 1,                                // ← bound
  "message": "breakpoint.provisionalBreakpoint"  // ← upstream text for this key is "Unbound breakpoint"
}
```

Two separate problems:

1. **Stale state (ours).** The provisional message is never cleared once the breakpoint verifies. `verified: true` + `adapterId: 1` alongside a message meaning "Unbound breakpoint" is self-contradictory, and `message` is exactly the field an agent reads to find out *why* a breakpoint is in a given state.
2. **Raw i18n key (upstream, but we ship it).** The string is a bundle key, not human text. In the vendored `vsDebugServer.js`, js-debug's pending-target stub answers `setBreakpoints` with `message: l10n.t("breakpoint.provisionalBreakpoint", "Unbound breakpoint")`, and since `L10N_FSPATH_TO_BUNDLE` is unset in our spawn the key is returned verbatim rather than the fallback.

## Reproduction (v0.24.2)

```jsonc
create_debug_session { language: "javascript" }
set_breakpoint { file: "examples/javascript/simple_test.js", line: 11,
                 logMessage: "LOGPOINT a={a}" }
start_debugging { scriptPath: "examples/javascript/simple_test.js",
                  dapLaunchArgs: { stopOnEntry: false } }
list_breakpoints { }
→ the record above
```

Reproduced on all three build groups (local, npx, docker) during a full `/testdebugger` sweep, then confirmed by hand. Plain line breakpoints show it too — it is not logpoint-specific.

## Severity

Low / cosmetic in isolation — `verified` and `adapterId` are correct, so nothing is functionally broken. Filing it because it is cheap to fix and it actively degrades the one free-text field an agent uses to diagnose breakpoint state: right now that field says "unbound" about a bound breakpoint, in a form no human or model can interpret.

## Suggested fix

Clear (or replace) the provisional `message` when the breakpoint transitions to `verified: true` — the same reconciliation path #439 added for forwarding the worker's initial `setBreakpoints` results would be the natural place. Optionally also map known js-debug l10n keys to their English fallbacks before surfacing them, so any other key that leaks reads as text.

Good first issue: small, well-isolated, easy to assert in a test (`list_breakpoints` on a verified JS breakpoint must not contain a `breakpoint.` key).

abhijeetnardele24-hash:
Closes #471. Clear or translate provisional breakpoint messages from js-debug (\reakpoint.provisionalBreakpoint\) when a breakpoint becomes verified.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,16 +4,6 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-overrides:
-  vite: 7.3.6
-  path-to-regexp: '>=8.4.0'
-  brace-expansion: '>=5.0.6'
-  hono: ^4.12.25
-  fast-uri: ^3.1.4
-  qs: ^6.15.2
-  ip-address: ^10.1.1
-  esbuild: ^0.28.2
-
 importers:
 
   .:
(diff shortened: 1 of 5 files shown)


debugmcpdev:
Thank you for this — you were first to pick up #471, and your instinct about the tricky part was genuinely right: the `else`-branch re-sanitizing the *stored* message when a verify update carries none is exactly the failure mode that a competing fix initially missed. Live verification confirmed js-debug's bind event omits `message`, and the merged fix (#476 + an amendment) handles it the same way you anticipated.

I'm closing this one in favor of #476, which arrived with unit tests and covered the `breakpoints_synced` reconciliation path as well. But please don't read that as "not wanted" — two specific things to fix in your workflow so your other PRs can land:

1. **The lockfile change is a security regression.** Your `pnpm-lock.yaml` was regenerated by a different pnpm version and silently **removed the entire `overrides` block** — the pinned security floors for `vite`, `qs`, `esbuild`, `hono`, and five others — and downgraded esbuild from 0.28.2 to 0.27.7. Any PR carrying that diff can't be merged regardless of the code around it. When `pnpm install` rewrites the lockfile on your machine, `git checkout pnpm-lock.yaml` before committing.
2. **`pnpm-workspace.yaml`** picked up literal placeholder text (`allowBuilds: esbuild: set this to true or false`) — worth double-checking `git diff` before pushing so tool-generated scaffolding doesn't ride along.

Root cause on our side: the repo didn't pin `packageManager`, so different pnpm majors disagree about the lockfile. A pin is landing shortly, which should make this whole class of problem disappear.

I've left detailed reviews on #474 and #475 — both are fixable and wanted. Thanks for contributing, and hope to merge those soon.

Issue #154, opened 2026-07-08
dev-proxy forwards backend stderr verbatim with naive chunk splitting


debugmcpdev:
## Problem

`tools/dev-proxy/dev-proxy.mjs` is outside the sanitization net that now covers the server and adapter packages (#146/#151/#153):

1. `logBackend()` splits each stderr/stdout chunk on `\n` without a persistent line buffer, so a line straddling two chunks is emitted as two partial `[backend]` lines — the same defect class fixed in #151.
2. No sanitization is applied before the lines are written to the dev proxy's own stderr.
3. `dev_rebuild_and_restart` embeds raw build stdout (truncated to 2000 chars) in the MCP tool response.

## Context

Dev-only tool, not published to npm, so exposure is limited to local development — filed for completeness from the #153 stderr audit. The backend's own logs are sanitized at their sources now, but anything the backend prints to raw stderr passes through verbatim.

## Suggested direction

`@debugmcp/shared` now exports `LineBuffer` and `sanitizeStderr` (moved there by the #153 fix). The dev proxy is plain .mjs but can import the built shared dist through the workspace link, or inline a minimal line buffer if keeping it dependency-free is preferred.

PR #474, opened 2026-08-24
Fix: Make dev-proxy build timeout configurable and distinct from failures

### LINKED ISSUE #472: dev-proxy: hardcoded 120s build timeout makes docker rebuilds report a successful build as 'Build failed'
## Summary

`dev_rebuild_and_restart` hardcodes a 120 s build timeout. A cold-cache `npm run docker-build` routinely exceeds that (the image is ~1.5 GB), and when the timeout trips the proxy reports **`Build failed:`** followed by a truncated tail of the Docker log — even though the build itself is fine. Re-running `npm run docker-build` by hand immediately afterwards succeeds and produces the image.

So the tool reports a successful build as a failure, and the accompanying output is a mid-`RUN` Dockerfile step rather than any diagnosis.

## Current behavior

```js
// tools/dev-proxy/dev-proxy.mjs:266
result = execSync(BUILD_CMD, {
  cwd: PROJECT_ROOT,
  encoding: 'utf-8',
  stdio: ['ignore', 'pipe', 'pipe'],
  timeout: 120000,          // hardcoded
  env: { ...process.env },
});
} catch (err) {
  const output = [err.stdout, err.stderr].filter(Boolean).join('\n') || err.message || String(err);
  throw new Error(`Build failed: ${sanitizeStderrTail(output, { maxLines: 20, maxChars: 2000 })}`);
}
```

`execSync` surfaces a timeout kill as a thrown error just like a non-zero exit, and the `catch` treats both identically. `err.killed` / `err.signal === 'SIGTERM'` are available and unused.

## Observed

```
dev_rebuild_and_restart  (mcp-debugger-docker)
→ { "success": false,
    "error": "Build failed: …cp -r /app/packages/adapter-rust/dist
               /app/node_modules/@debugmcp/adapter-rust/ (last 20 of 1454 lines)" }

$ npm run docker-build      # same command, run directly
… #69 naming to docker.io/library/mcp-debugger:local done
$ echo $?
0
```

Note the 1454-line build log truncated to its last 20 lines — for a timeout that tail is always an arbitrary mid-build step, never the reason.

## Requests

1. **Make the timeout configurable**, alongside the other knobs the proxy already exposes as env vars (`DEV_PROXY_PORT`, `DEV_PROXY_BUILD_CMD`, `DEV_PROXY_ROOT`, `DEV_PROXY_BACKEND_TRANSPORT`) — e.g. `DEV_PROXY_BUILD_TIMEOUT_MS`. A docker-group proxy could then be configured with a realistic ceiling.
2. **Raise the default**, or make it depend on the configured `BUILD_CMD`. 120 s is comfortable for `npm run build` but not for an image build.
3. **Distinguish a timeout from a build failure.** On `err.killed`/`SIGTERM`, say so — "Build timed out after 120000 ms (raise DEV_PROXY_BUILD_TIMEOUT_MS); the build may still have succeeded, re-run manually to confirm" — instead of `Build failed:`. This is the part that actually costs time: the message sends you looking for a build regression that isn't there.

## Impact

Dev tooling only (`tools/dev-proxy/`), no effect on the shipped server. Worth fixing because the docker group is unusable through `dev_rebuild_and_restart` on a cold cache, and the misreport is convincing enough to derail a debugging session — it did during a `/testdebugger` sweep, where the docker group had to be rebuilt out-of-band and restarted with `dev_restart_debugger`.

abhijeetnardele24-hash:
Closes #472. Adds DEV_PROXY_BUILD_TIMEOUT_MS and distinguishes SIGTERM/killed timeouts from real build failures in dev-proxy.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,16 +4,6 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-overrides:
-  vite: 7.3.6
-  path-to-regexp: '>=8.4.0'
-  brace-expansion: '>=5.0.6'
-  hono: ^4.12.25
-  fast-uri: ^3.1.4
-  qs: ^6.15.2
-  ip-address: ^10.1.1
-  esbuild: ^0.28.2
-
 importers:
 
   .:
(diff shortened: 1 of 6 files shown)


debugmcpdev (changes requested):
Thanks — the core change here is good and matches what #472 asked for: `DEV_PROXY_BUILD_TIMEOUT_MS` alongside the other proxy env knobs, and `err.killed || err.signal === 'SIGTERM'` to stop reporting a timed-out build as `Build failed:`. The timeout message text ("the build may still have succeeded, re-run manually to confirm") is exactly the guidance an agent needs at that moment. I'd like to merge this once the branch is cleaned up:

**Must fix**

1. **Rebase so this PR contains only the dev-proxy change.** The branch currently stacks on your #473 commit (that issue is now fixed via #476, so those files will conflict) and carries the regenerated `pnpm-lock.yaml` — which removes the repo's security `overrides` block (vite/qs/esbuild/hono/… floors) and downgrades esbuild — plus placeholder text in `pnpm-workspace.yaml`. Neither file should appear in the diff at all:
   ```
   git fetch upstream main
   git checkout -b fix/dev-proxy-timeout-v2 upstream/main
   git checkout fix/dev-proxy-timeout -- tools/dev-proxy/dev-proxy.mjs
   # then re-apply only the timeout hunks if the file picked up other edits
   ```
   (Or interactively rebase and drop everything except the `tools/dev-proxy/dev-proxy.mjs` hunks.)
2. **Restore the deleted comment block** above the `catch` (the one explaining why stderr is sanitized, referencing issue #154). It documents a real constraint — raw build stderr must not reach tool responses — and the constraint still applies to both of your new error paths.

**Should fix**

3. Guard against a malformed env var: `parseInt(...)` yields `NaN` for e.g. `DEV_PROXY_BUILD_TIMEOUT_MS=abc`, and `execSync` treats a NaN timeout as no timeout. Something like:
   ```js
   const parsed = parseInt(process.env.DEV_PROXY_BUILD_TIMEOUT_MS || '', 10);
   const BUILD_TIMEOUT_MS = Number.isFinite(parsed) && parsed > 0 ? parsed : 120000;
   ```
4. Document the new knob in the dev-proxy **Configuration** list in `CLAUDE.md` (it enumerates `DEV_PROXY_PORT`, `DEV_PROXY_BUILD_CMD`, …).

**Optional** — #472 also suggested raising the default when the build command is a docker build; fine to leave that out and let docker users set the env var, as long as the knob is documented.

No tests exist for `tools/dev-proxy/` today, so I won't ask for one — I'll verify the timeout path manually before merge. Happy to answer questions; once the branch is just the dev-proxy diff this should go in quickly.

debugmcpdev:
Closing in favor of #487, per your own note there — thanks for cutting the clean unstacked branch. One heads-up for future branches: this one's `pnpm-lock.yaml` was regenerated by a different pnpm major, which dropped the security `overrides` block (and `pnpm-workspace.yaml` picked up a placeholder `allowBuilds` entry). #479 pins `packageManager` to pnpm@10.33.0 so after it lands, `pnpm install` will use the right version automatically via corepack.

Issue #450, opened 2026-08-24
python attach has no path-mapping lever — localRoot/remoteRoot/pathMappings silently dropped


debugmcpdev:
## What

`attach_to_process` for python offers no way to map local source paths to debuggee paths. Any of `localRoot`/`remoteRoot`/`pathMappings` — passed top-level or via `adapterConfig` — are **silently dropped**: `transformAttachConfig` in `packages/adapter-python/src/python-debug-adapter.ts` builds a fresh whitelist object (`connect`/`justMyCode`/`stopOnEntry`/`cwd`/`env`) and the transform result replaces the merged config (`src/session/session-manager-operations.ts`, `transformedLaunchConfig ?? generic`).

Verified empirically during #424 (kind + EKS, 2026-08-23/24): against a pod running debugpy, a breakpoint at a local workstation path reports `verified: false` ("Breakpoint in file that does not exist") and never binds; the container-side path binds and hits.

## Why it matters

- Ruby has `localfsMap`; python — the most common attach target — has nothing. Agents debugging a pod must use debuggee-side paths (now documented in `docs/kubernetes.md` and `examples/kubernetes/attach-presets.md`), which works, but source-context tools on the client side can't correlate to the local checkout.
- The JIT-diagnostics tutorial shipped `localRoot`/`remoteRoot` in its Part 1 attach call for months — inert parameters that *looked* load-bearing. Fixed in #424's PR, but the silent drop is the underlying footgun: unknown attach keys should either work or warn.

## Proposal

1. Pass `pathMappings` through to debugpy: accept `adapterConfig.pathMappings` (debugpy's native `[{localRoot, remoteRoot}]` shape) in `transformAttachConfig` and forward it in the attach request. Optionally accept top-level `localRoot`/`remoteRoot` as sugar for a single mapping.
2. Consider a generic warning when adapter transforms drop non-reserved keys the caller explicitly provided (the #336 reserved-key strip already warns; this would extend the idea to whitelist-transform adapters).

## Acceptance

Attach to a pod's debugpy with `adapterConfig: {pathMappings: [{localRoot: "<abs local>", remoteRoot: "/app"}]}` → a breakpoint at the local path binds (`verified: true`) and hits; without mappings, current behavior unchanged.

PR #475, opened 2026-08-24
Fix: Explicit adapterConfig attach allow-lists and did-you-mean suggestions

### LINKED ISSUE #466: adapterConfig: unknown keys are accepted silently — a typo of a supported key (pathMapping vs pathMappings) gets no warning
`attach_to_process` warns about a short deny-list of known-bad `adapterConfig` keys, but any key outside that list is accepted in total silence. The most likely agent mistake — a typo of a *supported* key — therefore produces no feedback at all.

## Repro

Deny-listed key → warns correctly (this is #450 working as designed):

```jsonc
attach_to_process { port: 5679, adapterConfig: { localRoot: "/nope", … } }
→ { "success": true,
    "warning": "adapterConfig key(s) not supported by the python attach request were ignored: localRoot" }
```

Typo of the *correct* key → completely silent:

```jsonc
attach_to_process { port: 5691, adapterConfig: {
  pathMapping: [ { localRoot: "…/examples/python", remoteRoot: "…/examples/python" } ]
} }
→ { "success": true, "state": "paused" }        // no warning field at all
```

`pathMapping` vs `pathMappings`. One character. The agent now believes path mapping is configured; it isn't. In a real remote-attach (container debuggee, host checkout) the consequence is breakpoints that never bind, with the one lever that would have fixed it appearing — from the response — to have been accepted.

Same silence for an obviously-unknown key (`bogusAttachKey: 1`): accepted, unmentioned.

## Why the current design can't catch this

The check is a **deny-list** of enumerated ptvsd-era keys. By construction it only ever catches keys someone already thought to write down. Typos, stale keys from other debuggers, and keys borrowed from a different adapter's docs all sail through.

This is the same class of defect #450 was filed to remove — "unknown attach keys must either work or warn" — just reached through a different door. A typo'd key does neither observably: it doesn't work, and it doesn't warn.

## Suggested fix

Invert it: each adapter declares the `adapterConfig` keys it understands, and anything outside that set gets named in the existing `warning` field. That turns the check from "keys we remembered to ban" into "keys this adapter actually consumes", which catches typos for free.

Two refinements worth considering:
- **Did-you-mean.** With a known-good set in hand, `pathMapping` → `pathMappings` is a trivial edit-distance suggestion, and it is exactly the moment an agent can act on it.
- **Keep pass-through, just say so.** If forwarding unknown keys to the adapter is deliberate (some adapters accept extras the wrapper doesn't model), that's fine — but the response should still list what was forwarded-unrecognised, so silence never means "understood".

The `warning` plumbing already exists and is already surfaced top-level on the response; this is a change to *what populates it*, not new machinery.

Agent-ergonomics impact: this is a silent-wrong-state defect. The agent's next N steps are built on a false belief about the session's configuration, and nothing in any response contradicts it.

abhijeetnardele24-hash:
Closes #466. Adds \supportedAttachKeys\ to debug adapters to strictly warn on unknown adapterConfig keys during attach, replacing the previous deny-list behavior. Also includes a did-you-mean helper to provide typo suggestions (e.g. pathMapping -> pathMappings).

### DIFF
--- a/packages/adapter-javascript/src/javascript-debug-adapter.ts
+++ b/packages/adapter-javascript/src/javascript-debug-adapter.ts
@@ -621,6 +621,22 @@ export class JavascriptDebugAdapter extends EventEmitter implements IDebugAdapte
     return true;
   }
 
+  readonly supportedAttachKeys: readonly string[] = [
+    'stopOnEntry',
+    'justMyCode',
+    'timeout',
+    'cwd',
+    'env',
+    'sourceMaps',
+    'resolveSourceMapLocations',
+    'outFiles',
+    'localRoot',
+    'remoteRoot',
+    'smartStep',
+    'skipFiles',
+    'trace'
+  ];
+
   /**
    * Build the js-debug (pwa-node) attach configuration. Unlike
    * transformLaunchConfig — which always produces a launch request — this
--- a/packages/adapter-python/src/python-debug-adapter.ts
+++ b/packages/adapter-python/src/python-debug-adapter.ts
@@ -442,6 +442,23 @@ export class PythonDebugAdapter extends EventEmitter implements IDebugAdapter {
     return true;
   }
 
+  readonly supportedAttachKeys: readonly string[] = [
+    'pathMappings',
+    'subProcess',
+    'django',
+    'jinja',
+    'justMyCode',
+    'stopOnEntry',
+    'redirectOutput',
+    'showReturnValue',
+    'logToFile',
+    'sudo',
+    'args',
+    'cwd',
+    'env',
+    'envFile'
+  ];
+
   transformAttachConfig(config: GenericAttachConfig): PythonAttachConfig {
     const host = config.host || '127.0.0.1';
     const port = config.port;
--- a/packages/shared/src/interfaces/debug-adapter.ts
+++ b/packages/shared/src/interfaces/debug-adapter.ts
@@ -149,14 +149,19 @@ export interface IDebugAdapter extends EventEmitter {
   usesDirectConnectForAttach?(): boolean;
 
   /**
+   * The explicit set of adapterConfig keys this adapter understands for attach.
+   * Any keys not in this list will be surfaced as a warning to the user.
+   * If undefined, the adapter falls back to the old pass-through behavior.
+   */
+  readonly supportedAttachKeys?: readonly string[];
+
+  /**
    * Transform generic attach config to language-specific format
    * Only called if supportsAttach() returns true
    *
    * The result replaces the merged config verbatim as the DAP attach request
-   * body, so prefer a deny-list (strip known-bad keys, spread the rest) over
-   * an allowlist: unknown adapterConfig keys the caller provided should reach
-   * the debug adapter untouched. Caller-provided adapterConfig keys missing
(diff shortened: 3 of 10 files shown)


debugmcpdev (changes requested):
Thanks for taking on #466 — the direction is right (per-adapter declared key sets instead of a deny-list, plus did-you-mean; that's what the issue asked for), and the Levenshtein helper with the short-string threshold adjustment is a nice touch. It needs a round of changes before it can merge:

**Must fix**

1. **Un-stack the branch.** This PR currently contains your #473 and #474 commits plus the regenerated `pnpm-lock.yaml` (which removes the security `overrides` block and downgrades esbuild) and a placeholder edit to `pnpm-workspace.yaml`. Rebase onto upstream `main` with only the #466 changes; the lockfile and workspace files must not appear in the diff. (#471 is now fixed on main via #476, and #474 has its own review.)
2. **Truthful warning semantics.** For python, `transformAttachConfig` *spreads unknown keys through* to debugpy — so a key outside `supportedAttachKeys` is **forwarded unrecognised**, not "ignored". The issue was explicit about this refinement ("keep pass-through, just say so"): silence must never mean "understood", but the warning also must not claim we dropped something we actually sent. Suggest either (a) wording per adapter based on whether its transform passes unknowns through, or (b) actually stripping unknown keys when a `supportedAttachKeys` list exists, so "ignored" becomes true — (b) is cleaner but is a behavior change worth a sentence in the PR body.
3. **Keep stored state clean.** `session.attachDroppedConfigKeys` now stores decorated strings like `pathMapping (did you mean pathMappings?)`. That field is state, read later to build the user-facing warning — store bare key names and add the did-you-mean decoration only where the warning string is composed (`session-manager-operations.ts` ~line 2832). Otherwise any future consumer comparing those entries against real config keys breaks.
4. **Don't delete the #450 rationale.** The doc comment on `transformAttachConfig` in `packages/shared/src/interfaces/debug-adapter.ts` explains *why* the deny-list+pass-through contract existed. Amend it to describe the new `supportedAttachKeys` layer on top; the history of why is what stops the next person from re-simplifying it.
5. **Tests.** At minimum: unit tests for `didYouMean` (exact match, 1-edit typo, below/above threshold, short-string behavior), and an attach-warning test asserting the `pathMapping` → `pathMappings` suggestion end-to-end — `tests/core/unit/server/server-redefine-and-attach.test.ts` (the assertion at line ~441) shows the existing pattern for the #450 warning and must keep passing.

**Should fix**

6. Cite the source for each `supportedAttachKeys` list in a comment (debugpy's attach schema; js-debug's pwa-node attach options). These lists are now load-bearing for warnings — a reviewer needs to know they weren't guessed. If a key's status is unclear, leaving it off is fine since this is warn-only, but say so.

Adapters beyond python/javascript falling back to the old behavior is fine for this PR. This one isn't gating the 0.25.0 release, so there's no rush — but it's a wanted change and I'd like to land it with your name on it.

debugmcpdev:
Closing in favor of #488, per your own note there (review feedback is on #488). Same lockfile heads-up as #474: once #479 lands, the `packageManager` pin will keep local pnpm versions from regenerating `pnpm-lock.yaml` incompatibly.

Issue #439, opened 2026-08-23
list_breakpoints: logpoints (and any never-paused launch) stay verified:false with no adapterId after firing


debugmcpdev:
## Summary

For a launch that never pauses — the normal shape of a logpoint-only session — breakpoints stay `verified: false` with no `adapterId` in `list_breakpoints` forever, even though they demonstrably bound and fired. The post-launch re-sync added in #236 is gated on the session being `RUNNING`/`PAUSED`, and a short program is already terminal by the time the gate is evaluated.

## Reproduction

Rust, `examples/rust/hello_world`, logpoint on line 18 (the same line as a plain breakpoint that *does* verify — see contrast below):

```jsonc
// set_breakpoint { file: ".../src/main.rs", line: 18, logMessage: "LOGPOINT name={name}" }
// start_debugging { scriptPath: ".../target/debug/hello_world", dapLaunchArgs: { stopOnEntry: false } }
// -> state: "stopped" (ran to completion without pausing, as logpoints should)

// get_output -> the logpoint FIRED and interpolated correctly:
{ "category": "console", "output": "LOGPOINT name=\"Rust\"\n" }

// list_breakpoints ->
{
  "id": "df2ede57-...", "file": ".../src/main.rs", "line": 18,
  "logMessage": "LOGPOINT name={name}",
  "verified": false,          // <-- wrong: it bound and fired
  "requestedLine": 18
                              // <-- no adapterId, no "Resolved locations" message
}
```

**Not language-specific** — identical result with Python (`examples/python/simple_test.py:11`, `logMessage: "LOGPOINT a={a}"`): logpoint fires, `list_breakpoints` reports `verified: false`, no `adapterId`.

**Contrast — the same rust file/line as a plain breakpoint** (session pauses, so the re-sync runs):

```jsonc
{ "verified": true, "message": "Resolved locations: 1", "adapterId": 1 }
```

Also visible after a normal breakpoint session runs to completion — the state is correct there only because the re-sync already ran while paused.

## Root cause

`src/session/session-manager-operations.ts` — the worker applies `initialBreakpoints` itself and its `setBreakpoints` responses never reach the session store, so the store depends entirely on this re-sync:

```ts
if (finalSession.breakpoints.size > 0 &&
    (finalState === SessionState.RUNNING || finalState === SessionState.PAUSED)) {
  for (const file of files) await this.syncBreakpointsForFile(finalSession, file);
}
```

[...]

PR #476, opened 2026-08-24
fix(breakpoints): clear the provisional message once a breakpoint verifies

### LINKED ISSUE #471: Verified JS breakpoints keep the raw provisional message 'breakpoint.provisionalBreakpoint' (untranslated key + contradicts verified:true)

MsfPablo:
Closes #471.

Both halves of the issue, fixed at the point where an adapter message is stamped onto a stored breakpoint.

**1. Stale state.** The provisional message was never cleared once the breakpoint verified, so `list_breakpoints` asserted `verified: true` + `adapterId` alongside a message meaning "Unbound breakpoint".

**2. Raw l10n key.** The string arrived as a bundle key rather than text, since `L10N_FSPATH_TO_BUNDLE` is unset in the adapter spawn and `l10n.t(key, fallback)` then returns the key verbatim.

`normalizeBreakpointMessage(message, verified)` in `src/utils/breakpoint-message.ts` drops a provisional note when `verified` is true, and otherwise resolves known js-debug keys to the English fallback declared at their own `l10n.t` call site. So the field either says something true or says nothing — and a leaked key never reaches the user either way.

**Applied at both stamp sites**, not just one. The issue points at the #439 reconciliation path, but that is only half of it: the comment at `session-manager-core.ts:823` notes js-debug's worker path never emits `breakpoints_synced`, so for JavaScript the message actually lands via the `breakpoint` event handler. Fixing only the #439 path would have left the reported repro untouched. Both now go through the same helper.

**Test** — `tests/unit/utils/breakpoint-message.test.ts`, including the assertion the issue suggests (a verified breakpoint must not carry a `breakpoint.` key):

```
$ npx vitest run tests/unit/utils/breakpoint-message.test.ts   # 5 passed
$ npx tsc --noEmit -p tsconfig.json                            # clean
$ npx vitest run tests/unit tests/core/unit                    # 3205 passed
```

Against `main` the same suites give 3200 passed with an identical 3 failures, all in the dotnet adapter and unrelated to this change — so this PR adds 5 passing tests and no failures.

**Two environment notes, neither caused by this PR:** `pnpm install --frozen-lockfile` fails on `main` with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` (overrides drift), and the `postinstall` CodeLLDB vendor step fails to fetch on this platform. I installed with `--no-frozen-lockfile` and deliberately reverted `pnpm-lock.yaml` so it is not part of the diff. Worth a look separately if you weren't aware.

I did not touch the vendored `vsDebugServer.js` — setting `L10N_FSPATH_TO_BUNDLE` at spawn time would fix the key at the source for *all* messages, but that is a larger change to adapter spawning and felt out of scope here. Happy to look at it if you'd prefer that direction.

### DIFF
--- a/src/session/session-manager-core.ts
+++ b/src/session/session-manager-core.ts
@@ -615,11 +615,15 @@ export abstract class SessionManagerCore extends EventEmitter {
       if (typeof eventBp.line === 'number') {
         target.line = eventBp.line;
       }
-      if (eventBp.message !== undefined) {
-        // A provisional "unbound" note must not survive verification, and a
-        // leaked l10n key must not reach the user (issue #471).
-        target.message = normalizeBreakpointMessage(eventBp.message, target.verified);
-      }
+      // A provisional "unbound" note must not survive verification, and a
+      // leaked l10n key must not reach the user (issue #471). When the event
+      // carries no message, re-normalize the stored one: js-debug's bind event
+      // omits `message`, so a provisional note stamped earlier would otherwise
+      // outlive the verification it contradicts.
+      target.message = normalizeBreakpointMessage(
+        eventBp.message !== undefined ? eventBp.message : target.message,
+        target.verified
+      );
       if (typeof eventBp.id === 'number') {
         target.adapterId = eventBp.id;
       }
@@ -855,8 +859,13 @@ export abstract class SessionManagerCore extends EventEmitter {
         }
         // Stamp message only when present — a clean sync must not wipe the
         // capability-drift warning handleAdapterCapabilities may have set.
+        // (Normalization drops only provisional "unbound" notes, so a sync
+        // that verifies without a message clears a stale one while other
+        // stored messages pass through untouched — issue #471.)
         if (result.message !== undefined) {
           target.message = normalizeBreakpointMessage(result.message, target.verified);
+        } else if (target.message !== undefined && target.verified) {
+          target.message = normalizeBreakpointMessage(target.message, target.verified);
         }
         this.logger.info('debug:breakpoint', {
           event: 'verified',
--- a/src/session/session-manager-core.ts
+++ b/src/session/session-manager-core.ts
@@ -24,6 +24,7 @@ import { ISessionStoreFactory } from '../factories/session-store-factory.js';
 import { IProxyManager } from '../proxy/proxy-manager.js';
 import { IProxyManagerFactory } from '../factories/proxy-manager-factory.js';
 import type { BreakpointSyncResult, FunctionBreakpointSyncResult } from '../proxy/dap-proxy-interfaces.js';
+import { normalizeBreakpointMessage } from '../utils/breakpoint-message.js';
 import { IAdapterRegistry } from '@debugmcp/shared';
 
 // Custom launch arguments interface extending DebugProtocol.LaunchRequestArguments
@@ -615,7 +616,9 @@ export abstract class SessionManagerCore extends EventEmitter {
         target.line = eventBp.line;
       }
       if (eventBp.message !== undefined) {
-        target.message = eventBp.message;
+        // A provisional "unbound" note must not survive verification, and a
+        // leaked l10n key must not reach the user (issue #471).
+        target.message = normalizeBreakpointMessage(eventBp.message, target.verified);
       }
       if (typeof eventBp.id === 'number') {
         target.adapterId = eventBp.id;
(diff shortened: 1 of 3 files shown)


debugmcpdev:
Thank you — this is an outstanding first contribution. The analysis was exactly right on both counts: the two stamp sites (the #439 reconciliation path *and* the breakpoint-event path js-debug actually uses), and the l10n-key mechanism. The PR body alone saved a lot of review time.

Before merging I ran the literal #471 repro against your branch, and it surfaced one subtle gap I've fixed with a commit pushed to your branch (15af266a — hope you don't mind, we're bundling this for the 0.25.0 release):

- The provisional message is stamped **while the breakpoint is still unverified**, so `normalizeBreakpointMessage` stores it in translated form (`"Unbound breakpoint"`).
- js-debug's later bind event flips `verified: true` **without a `message` field**, so the `if (eventBp.message !== undefined)` guard never revisited the stored text.
- Net effect on your branch: the raw key was gone (good), but `list_breakpoints` still reported `verified: true` + `message: "Unbound breakpoint"`.

The amendment keeps your helper and applies it slightly more aggressively: the translated fallback text counts as provisional too, and both stamp sites re-normalize the *stored* message when an update carries none. Re-verified against the repro: `verified: true`, `adapterId` set, and no `message` field at all. Your 5 tests still pass, plus 2 new ones for the translated form. Full unit suites: 3210 passed.

On your two environment notes — both appreciated:
1. `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` — root cause is that the repo doesn't pin `packageManager`, so a different pnpm major computes the lockfile settings differently. A pin + CONTRIBUTING note is landing shortly. Reverting the lockfile out of your diff was exactly the right call.
2. The CodeLLDB postinstall fetch failure — could you share your platform and the error output? Happy to file and fix that separately.

Will merge once CI is green. Thanks again — contributions at this level of care are very welcome here.

debugmcpdev (approved):

PR #477, opened 2026-08-24
Fix: build-npx-tarball.cjs skips prepare-pack

### LINKED ISSUE #463: build-npx-tarball.cjs skips prepare-pack — dist-tarball ships workspace:* and cannot be npm install'ed
`scripts/build-npx-tarball.cjs` produces `dist-tarball/mcp-debugger-latest.tgz` by calling `npm pack` directly, without the `prepare-pack.js prepare` / `restore` pair. It therefore captures `packages/mcp-debugger/package.json` **unrewritten**, so the tarball ships pnpm-only `workspace:*` specifiers and cannot be installed by npm.

## Repro

```console
$ node scripts/build-npx-tarball.cjs
$ mkdir /tmp/iso && cd /tmp/iso && npm init -y
$ npm install /path/to/repo/dist-tarball/mcp-debugger-latest.tgz
npm error code EUNSUPPORTEDPROTOCOL
npm error Unsupported URL Type "workspace:": workspace:*
```

The manifest inside the tarball:

```jsonc
"optionalDependencies": {
  "@debugmcp/codelldb-darwin-arm64": "workspace:*",
  "@debugmcp/codelldb-darwin-x64":   "workspace:*",
  "@debugmcp/codelldb-linux-arm64":  "workspace:*",
  "@debugmcp/codelldb-linux-x64":    "workspace:*",
  "@debugmcp/codelldb-win32-x64":    "workspace:*"
}
```

## The two pack paths disagree

| Path | Output | `optionalDependencies` |
|---|---|---|
| `packages/mcp-debugger/scripts/bundle-cli.js` (calls `prepare-pack.js`) | `packages/mcp-debugger/package/*.tgz` | `1.11.8` ✅ |
| `scripts/build-npx-tarball.cjs` (no prepare-pack) | `dist-tarball/mcp-debugger-latest.tgz` | `workspace:*` ❌ |

`prepare-pack.js` itself is fine — it already covers `optionalDependencies` and lists all five `codelldb-*` workspaces. It just isn't invoked on this path. Note `bundle-cli.js` runs prepare-pack during `npm run build` and then **restores** the original manifest, so by the time `build-npx-tarball.cjs:43` runs `npm pack`, `package.json` is back to `workspace:*`.

## Why it matters

The **release** path is unaffected — releases pack via `bundle-cli.js`. Two consequences that do bite:

1. `dist-tarball/mcp-debugger-latest.tgz` is what the documented npx dev-proxy config consumes, and it can't be `npm install`ed outside the repo.
2. It makes the npx test arm **not packaging-faithful**. `build-npx-tarball.cjs` extracts to `dist-tarball/extracted/` *inside the repo*, so CodeLLDB resolution finds the repo's own vendor tree (resolution order is vendor → `CODELLDB_PATH` → platform package). `doctor --json` consequently reports `codelldbSource: "vendored"` on npx where the packaging assertion expects `platform-package` — so the sweep's sharpest npx packaging check silently can't fail.

(Independently: `@debugmcp/codelldb-win32-x64` is still E404 on the registry, so `platform-package` is unreachable until those publish — but that's the known v0.25.0 hold, not this bug.)

## Fix

Wrap the pack in `scripts/build-npx-tarball.cjs` (~line 43):

```js
run('node', ['scripts/prepare-pack.js', 'prepare']);
try {
  run('npm', ['pack', '--pack-destination', TARBALL_DIR], { cwd: PKG_DIR });
} finally {
  run('node', ['scripts/prepare-pack.js', 'restore']);
}
```

Worth a regression test that greps the packed manifest for `workspace:` and fails — it's a one-line assertion that would have caught this and would also guard the release path.

abhijeetnardele24-hash:
Closes #463. Wraps the pack command in build-npx-tarball.cjs with prepare-pack prepare and restore so the tarball contains correct package version specifiers.

### DIFF
--- a/scripts/build-npx-tarball.cjs
+++ b/scripts/build-npx-tarball.cjs
@@ -40,11 +40,20 @@ fs.mkdirSync(TARBALL_DIR, { recursive: true });
 run('pnpm', ['install', '--frozen-lockfile=false']);
 run('npm', ['run', 'build']);
 run('npm', ['run', 'build:packages']);
+// Rewrite workspace:* specifiers to real versions before packing (issue #463)
+// — npm cannot install a tarball that ships pnpm-only workspace: protocol.
 run('node', ['scripts/prepare-pack.js', 'prepare']);
-try {
-  run('npm', ['pack', '--pack-destination', TARBALL_DIR], { cwd: PKG_DIR });
-} finally {
-  run('node', ['scripts/prepare-pack.js', 'restore']);
+// run() exits the process on failure, which would skip a `finally` — spawn
+// the pack directly so the original package.json is restored either way.
+process.stderr.write(`[build-npx-tarball] npm pack --pack-destination ${TARBALL_DIR}\n`);
+const packResult = spawnSync('npm', ['pack', '--pack-destination', TARBALL_DIR], {
+  stdio: 'inherit',
+  shell: process.platform === 'win32',
+  cwd: PKG_DIR,
+});
+run('node', ['scripts/prepare-pack.js', 'restore']);
+if (packResult.status !== 0) {
+  process.exit(typeof packResult.status === 'number' ? packResult.status : 1);
 }
 
 const candidates = fs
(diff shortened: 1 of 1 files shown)


debugmcpdev:
This one is exactly right — clean branch, single concern, and the prepare/restore wrapping matches what the issue called for. Thank you, this is release-blocking for 0.25.0 so the turnaround is much appreciated.

I pushed one hardening commit to your branch (ea804a40) rather than round-tripping, since we're bundling for the release:

1. `run()` calls `process.exit` on failure, which skips `finally` — so a failed `npm pack` would have left `packages/mcp-debugger/package.json` in its rewritten state (and the next `prepare` would then back up the wrong file). The pack is now spawned directly and restore runs on both paths.
2. Added the regression guard the issue asked for: after extraction, the build fails if the packed manifest still contains `workspace:` specifiers — so this path can't silently regress again.

Verified end to end on your branch: the packed `optionalDependencies` now carry real versions (`1.11.8`), `npm install <tarball>` succeeds in an isolated directory (previously `EUNSUPPORTEDPROTOCOL`), and the installed CLI runs. Will merge once CI is green.

Also: this branch is exactly the shape we want PRs in — no lockfile churn, no unrelated commits. If you bring #474 and #475 to the same shape per the reviews there, those can land too.

debugmcpdev (approved):
Verified end to end (see comment): packed manifest carries real versions, tarball npm-installs in isolation, CLI runs. Hardening commit adds restore-on-failure and the workspace: regression guard.

PR #487, opened 2026-08-25
Fix: dev-proxy timeout configuration

### LINKED ISSUE #472: dev-proxy: hardcoded 120s build timeout makes docker rebuilds report a successful build as 'Build failed'

abhijeetnardele24-hash:
Closes #472 (Replacing #474 with a clean, unstacked branch). Adds DEV_PROXY_BUILD_TIMEOUT_MS and correctly intercepts SIGTERM/timeouts so they are distinct from actual build failures.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -444,8 +444,10 @@ The dev proxy (`tools/dev-proxy/dev-proxy.mjs`) is a lightweight MCP proxy that 
 **Configuration** (env vars, all optional):
 - `DEV_PROXY_PORT` — Backend HTTP port (default: 3001; used by `http` and `sse` modes)
 - `DEV_PROXY_BUILD_CMD` — Build command (default: `npm run build`)
+- `DEV_PROXY_BUILD_TIMEOUT_MS` — Build timeout in milliseconds (default: 120000)
 - `DEV_PROXY_ROOT` — Project root (default: auto-detected)
 - `DEV_PROXY_BACKEND_TRANSPORT` — `http` (default), `sse` (legacy/deprecated), or `stdio`
+- `DEV_PROXY_BACKEND_CMD` — Custom backend command override (e.g. `docker run ...`)
 
 #### Verify Installation
 
--- a/tools/dev-proxy/dev-proxy.mjs
+++ b/tools/dev-proxy/dev-proxy.mjs
@@ -14,6 +14,7 @@
  * Configuration (all env vars, all optional):
  *   DEV_PROXY_PORT               - Backend HTTP port (default: 3001, http/sse modes only)
  *   DEV_PROXY_BUILD_CMD          - Build command (default: "npm run build")
+ *   DEV_PROXY_BUILD_TIMEOUT_MS   - Build timeout in milliseconds (default: 120000)
  *   DEV_PROXY_ROOT               - Project root (default: auto-detected)
  *   DEV_PROXY_BACKEND_TRANSPORT  - "http" (default), "sse" (legacy), or "stdio"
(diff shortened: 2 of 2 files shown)


debugmcpdev:
Merging — thanks for the clean v2, and for validating the env var input (non-finite/negative falls back to the default, nice touch). One non-blocking follow-up idea: naming `DEV_PROXY_BUILD_TIMEOUT_MS` in the timeout message itself would tell the reader which knob to raise without opening the docs. Credited in the CHANGELOG under [Unreleased].

PR #540, opened 2026-08-28
docs: remove stale tool count from dev_server_status description

### LINKED ISSUE #539: chore(dev-proxy): dev_server_status still advertises a 'tool count' it never returns (2 spots left after #537)
## Summary

`dev_server_status` advertises a **tool count** that it has never returned. PR #537 corrected this in
`tools/dev-proxy/README.md`, but two other copies of the same wrong claim survived — including the one
an agent actually reads at call time.

## Ground truth

`BackendManager.getStatus()` (`tools/dev-proxy/dev-proxy.mjs:304-320`) returns exactly:

```js
{ state, pid, port, uptime, projectRoot, buildCmd, backendTransport, backendCmd }
```

No tool count. There is no field it could plausibly be a stale name for.

## Remaining stale copies

- `tools/dev-proxy/dev-proxy.mjs:614` — the MCP tool `description` string:
  `'Get the current status of the mcp-debugger backend (state, PID, uptime, tool count, port).'`
- `CLAUDE.md:504` — `` `dev_server_status` - Check backend state, PID, uptime, tool count ``

Already correct after #537, for reference:

- `tools/dev-proxy/README.md:86` — `Check backend state, PID, uptime, transport, project root, and port.`

## Why this one matters slightly more than a README typo

The `dev-proxy.mjs:614` string is the tool description served over MCP, so it is what an LLM sees when
deciding whether to call `dev_server_status`. Advertising a field the response never contains invites a
call made specifically to read a tool count, followed by a confused re-read of the payload. It is small,
but it is the copy with a live consumer.

## Suggested fix

Align both with the README wording:

> Get the current status of the mcp-debugger backend (state, PID, uptime, transport, project root, port).

## Why this is filed rather than just pushed

`main` requires a PR plus a full status-check run, so a standalone two-line docs change costs an entire
CI cycle. Not worth it alone — **please fold this into the next PR that already touches the dev proxy or
CLAUDE.md.** No behavior change, no tests needed.

Newer1107:
## Summary

Removes the stale "tool count" reference from the `dev_server_status` MCP tool description and the CLAUDE.md reference.

PR #537 corrected the README, but two other copies survived — including the one an LLM reads at call time.

## Changes

- `tools/dev-proxy/dev-proxy.mjs:614`: Updated MCP tool description to match README wording
- `CLAUDE.md:504`: Updated `dev_server_status` reference to match README wording

## Related Issues

Resolves #539

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -501,7 +501,7 @@ Each session also exposes its captured output as an MCP resource (`debug://sessi
 **Dev proxy only** (these 3 tools are injected by the dev proxy process itself, not by the main mcp-debugger server):
 - `dev_restart_debugger` - Restart the backend (pass `rebuild: true` to build first)
 - `dev_rebuild_and_restart` - Run `npm run build` then restart the backend
-- `dev_server_status` - Check backend state, PID, uptime, tool count
+- `dev_server_status` - Check backend state, PID, uptime, transport, project root, and port
 
 ### Troubleshooting MCP Connection
 - **If server shows "Failed to connect"**:
--- a/tools/dev-proxy/dev-proxy.mjs
+++ b/tools/dev-proxy/dev-proxy.mjs
@@ -611,7 +611,7 @@ const DEV_TOOLS = [
   {
     name: 'dev_server_status',
     description:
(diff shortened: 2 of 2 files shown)


debugmcpdev:
Merged — thanks, and welcome!

Issue #468, opened 2026-08-24
get_local_variables anchors to stackFrames[0] — returns empty on a syscall pause after already fetching the user frame's locals


debugmcpdev:
## Summary

`get_local_variables` hard-anchors to `stackFrames[0]`. When a `pause_execution` lands inside a blocking syscall — the normal case for any long-running program — frame 0 is a stdlib/libc frame with no locals, so the tool returns an empty array even though the user frame is **one frame down and already visible in the default (filtered) stack**.

The sting: the implementation has *already fetched* the data it then discards. `session-manager-data.ts` loops over **every** frame collecting scopes (the fan-out from #438), and then narrows the result to `stackFrames[0]`'s Local scope.

```ts
// src/session/session-manager-data.ts:436
const topFrame = stackFrames[0];
...
// the loop directly below fetches scopes for ALL frames
for (const frame of stackFrames) { const scopes = await this.getScopes(sessionId, frame.id); ... }
```

This matters because `get_local_variables` is the tool an agent reaches for to inspect a paused long-running process, and its documented purpose is "returns just the local variables without needing to traverse stack->scopes->variables manually". In exactly that scenario it returns nothing, and the agent has no signal that the data is one hop away.

## Reproduction (C/C++, verified on v0.24.2)

```jsonc
create_debug_session { language: "cpp" }
start_debugging { scriptPath: "examples/cpp/pause_test.cpp", dapLaunchArgs: { stopOnEntry: false } }
pause_execution { }
→ { stopReason: "pause", location: { file: ".../this_thread_sleep.h", line: 80 } }

get_local_variables { }
→ { "variables": [], "count": 0,
    "frame": { "name": "void std::this_thread::sleep_for<...>(...)",
               "file": "/usr/include/c++/13/bits/this_thread_sleep.h", "line": 80 },
    "message": "The Local scope is empty." }

get_stack_trace { }                       // DEFAULT call — already filtered
→ stackFrames[0] = std::this_thread::sleep_for   (this_thread_sleep.h:80)
  stackFrames[1] = main                          (pause_test.cpp:20)   ← VISIBLE, not hidden
  count: 5, hiddenFrames: 2

get_scopes { frameId: <main> }  → Local scope 1074
get_variables { scope: 1074 }   → [ { "name": "counter", "value": "348", "type": "long long" } ]
```

`main` is **frame 1 of the default, already-filtered stack** — not something `includeInternals` was hiding.

## Cross-adapter inconsistency

[...]

Issue #438, opened 2026-08-23
get_local_variables: truncation summary counts variables from discarded scopes, not the returned payload


debugmcpdev:
## Summary

`get_local_variables` reports a `truncation` summary whose counts describe variables that were **fetched and then discarded**, not the variables actually returned. The result is a response that says dozens of values were cut when none of the returned values were touched.

Follow-up defect in the size guard shipped by #359.

## Reproduction

Local (stdio/http) server, JavaScript, paused at `examples/javascript/simple_test.js:14`:

```jsonc
// get_local_variables { sessionId }
{
  "variables": [
    { "name": "a", "value": "1", "type": "number" },   // not truncated
    { "name": "b", "value": "2", "type": "number" }    // not truncated
  ],
  "count": 2,
  "truncation": {
    "omittedCount": 0,
    "valueTruncatedCount": 78,
    "notice": "Response size-guarded: 78 value(s) cut at 1024 chars (flagged truncated:true). Pass names: [\"a\",\"b\"] to fetch specific variables in full, or ..."
  }
}
```

The response contains 2 variables, both complete, **zero** flagged `truncated: true` — while the notice claims 78 values were cut and flagged.

Observed on every JavaScript run in a 48-combo sweep across all three deployment groups (local/docker/npx, primary and stdio variants), and also on Ruby (`valueTruncatedCount` 4–8 with 2 short values returned). Any language whose stack exposes large non-local scopes can hit it.

## Root cause

`getLocalVariables` (`src/session/session-manager-data.ts`) fans out over **every scope of every frame** on the stack to build `variablesMap`, pushing each fetch's truncation summary into `truncationSummaries`:

```ts
for (const frame of stackFrames) {
  for (const scope of scopes) {
    const detailed = await this.getVariablesDetailed(sessionId, scope.variablesReference, names);
    truncationSummaries.push(detailed.truncation);   // <-- every scope, incl. Global/Closure
    ...
  }
}
```

The returned payload is then narrowed to a single scope by `policy.extractLocalVariables(...)` (plus the `names` filter), but the reported summary still merges **all** the fan-out summaries:

```ts
const cappedLocals = applyVariableCaps(localVars);
const truncation = mergeTruncationSummaries([
  ...truncationSummaries,      // <-- counts from scopes that were discarded
  cappedLocals.truncation,
  ...
]);
```

So in JS the 78 cut values come from Global/Closure scopes that never reach the caller.

## Impact

[...]

PR #554, opened 2026-08-28
fix(inspection): improve local and thread fallbacks

### LINKED ISSUE #548: get_local_variables returns empty on a JS pause when the top frame's Local scope is empty but its Module/closure scope holds the value (#468 gap)
## Summary

The `#468` walk-down re-anchors to a lower **frame** when the top frame yields no locals. It does not consider the top frame's **other scopes**. On JavaScript this leaves `get_local_variables` returning an empty array at a perfectly ordinary pause, while the value the caller wants is one `get_variables` call away on the *same* frame.

## Repro

Pause fixture `examples/javascript/pause_test.js`:

```js
let counter = 0;
setInterval(() => {
  counter += 1;     // <- line 4
}, 500);
```

1. `create_debug_session {language: "javascript"}`
2. `start_debugging {scriptPath: ".../examples/javascript/pause_test.js", dapLaunchArgs: {stopOnEntry: false}}`, wait 3s
3. `pause_execution` — lands on `pause_test.js:4`, the `setInterval` arrow body (genuine **user** code, not a runtime frame)
4. `get_local_variables`

Actual:

```json
{"variables": [], "count": 0, "scopeName": "Local", "message": "The Local scope is empty."}
```

No `note`, no re-anchor.

But on that **same** frame:

- `get_scopes {frameId: 978}` → `Local` (ref 1, genuinely empty — the arrow function has no own locals), `Module` (ref 2), `Global` (ref 3)
- `get_variables {scope: 2}` → `counter = 46`

## Why the walk-down misses it

The heuristic descends to a lower *frame*. Here `stackFrames[0]` is legitimate user code, so there is nothing to walk down to that helps (`get_scopes` on the frame below returns `[]`); the miss is that the fallback stops at an empty `Local` scope instead of falling through to the sibling `Module` scope on the frame it already selected.

The tool clearly *can* report `Module` — during the 24-step breakpoint workflow, step 17 (after `step_out`) returned `scopeName: "Module"`. So this is a fallback-ordering gap, not a missing capability.

## Scope of the failure

Reproduced on **every** JS build/transport in the 2026-08-28 testdebugger run — 6/6:

| Server | Group/transport | P6 |
|---|---|---|
| mcp-debugger | local / http | FAIL (empty) |
| mcp-debugger-stdio | local / stdio | FAIL (empty) |
| mcp-debugger-npx | npx / sse | FAIL (empty) |
| mcp-debugger-npx-stdio | npx / stdio | FAIL (empty) |
| mcp-debugger-docker | docker / sse | FAIL (empty) |
| mcp-debugger-docker-stdio | docker / stdio | FAIL (empty) |

Not transport-specific — it is in the shared anchor-selection logic.

## Suggested fix

When the anchored frame's chosen scope is empty, try that frame's remaining scopes (`Module`/`Closure`/`Block`, ahead of `Global`) *before* walking to a lower frame, and disclose whichever it used in `note`/`scopeName` the same way the frame walk-down already does.

## Not a #496 regression

Checked separately: the anchor adopted by `get_stack_trace {threadId}` survives an intervening `list_threads` (P7c). The empty locals there are this same issue, not an anchor revert.

Found by the `/testdebugger` suite, 2026-08-28. Related: #468 (the frame walk-down this extends), and the Ruby sibling case filed alongside.

### LINKED ISSUE #549: Ruby: rdbg's %self pseudo-variable defeats the #468 walk-down — get_local_variables returns only %self at a sleep-blocked pause
## Summary

The `#468` walk-down re-anchors when the top frame "yields no locals". Its emptiness test appears to be *scope is non-empty*. rdbg injects a synthetic `%self` entry into **every** frame's local-variables scope — including native `[C]` frames — so a frameless-in-practice `[C] Kernel#sleep` frame looks like it yields locals, the walk-down never fires, and the caller gets a scope with nothing useful in it.

The user-visible symptom is exactly the one `#468` was opened to fix, just for Ruby.

## Repro

Pause fixture `examples/ruby/long_running.rb` (counter loop with `sleep 1`).

1. `create_debug_session {language: "ruby"}`
2. `start_debugging {scriptPath: ".../examples/ruby/long_running.rb", dapLaunchArgs: {stopOnEntry: false}}`, wait 3s
3. `pause_execution` — lands with `stackFrames[0] = [C] Kernel#sleep`
4. `get_local_variables`

Actual:

```json
{"frame": {"name": "[C] Kernel#sleep", "line": 13},
 "count": 1,
 "variables": [{"name": "%self", "value": "main"}],
 "scopeName": "Local variables"}
```

— and **no `note`**, because no walk-down happened.

Walking down one frame by hand (`get_scopes` on the next frame, `block in <main>`, then `get_variables`) returns the real locals:

```
counter = 37
squared = 1369
message = "tick"
```

So the data is one frame away and reachable; the heuristic simply stops too early.

## Contrast with the adapters where #468 works

On the same run, cpp / rust / .NET / java all re-anchored correctly and disclosed it, e.g. cpp:

```
note: Top frame 'clock_nanosleep' has no local variables (runtime/stdlib frame); showing frame #3 'main' instead
counter = 373
```

Ruby is the outlier purely because its "empty" scope is never literally empty.

## Scope of the failure

Reproduced on 3/3 ruby-launch-capable servers in the 2026-08-28 testdebugger run:

| Server | Group/transport | P6 |
|---|---|---|
| mcp-debugger | local / http | FAIL (`%self` only) |
| mcp-debugger-stdio | local / stdio | FAIL (`%self` only) |
| mcp-debugger-npx | npx / sse | FAIL (`%self` only) |

(Docker excluded — ruby launch is unavailable in the image.)

## Suggested fix

In the walk-down's "does this frame yield locals?" predicate, treat a scope whose only entries are adapter-synthetic pseudo-variables as empty. For Ruby that is `%self` (and any other `%`-prefixed rdbg pseudo-entries). Ideally the policy layer supplies the ignore-list so the core predicate stays adapter-agnostic.

An alternative/complementary signal: `[C] `-prefixed frame names are rdbg's marker for native frames, which never carry Ruby locals.

## Everything else in the ruby run was clean

No `#492` proxy-init timeout, no `#470` framing errors, `#496` anchor durability held, and the ruby launch-config workflow (`dapLaunchArgs.env` #318, `.cwd` #320, mid-run stdout streaming #317) passed 7/7.

Found by the `/testdebugger` suite, 2026-08-28. Related: #468, and the JavaScript sibling-scope case filed alongside (#548).

### LINKED ISSUE #553: get_stack_trace: explicit threadId with no frames returns a silent empty result (no note, lastStop.threadId names a different thread)
## Summary

`get_stack_trace` with an explicit `threadId` that has no frames returns a silent empty result:

```json
{"success":true,"stackFrames":[],"count":0,"includeInternals":false,
 "stopReason":"pause","lastStop":{"reason":"pause","threadId":1,...}}
```

No `note`, no `hiddenFrames`, and `lastStop.threadId` still names a *different* thread (1), so nothing in the response tells the caller that the thread they asked about is the one with no frames — or what to do next.

This is the one empty-stack path that still answers silently. Elsewhere the server is careful about this: the tool description promises a `note` when frames are hidden, and `threadId` is documented as *"the escape hatch when the session is anchored to a frameless thread."* When the escape hatch itself comes up empty, the agent is left with no signal at all.

## Reproduction

Java attach to a plain running JVM (no `suspend=y` needed):

```bash
javac -g -d examples/java examples/java/PauseTest.java
java -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5007 -cp examples/java PauseTest
```

```jsonc
create_debug_session { "language": "java" }
attach_to_process    { "sessionId": "...", "port": 5007 }
list_threads         { "sessionId": "..." }
// → 1 main, 2 Reference Handler, 3 Finalizer, 4 Signal Dispatcher,
//   5 Attach Listener, 6 Notification Thread, 7 Common-Cleaner

get_stack_trace { "sessionId": "...", "threadId": 4 }   // Signal Dispatcher
get_stack_trace { "sessionId": "...", "threadId": 6 }   // Notification Thread
```

Both return `stackFrames: []`, `count: 0`, **no `note`**. Reproduced on local `dist/` @ 002a49d2, Windows 11, JDK 21.0.10.

## This is not the internal-frame filter

Worth ruling out, since a filter that ate every frame would be a different (worse) bug:

| call | result |
|---|---|
| `threadId: 4` | `count: 0` |
| `threadId: 4, includeInternals: true` | `count: 0` — still empty |
| `threadId: 3` (Finalizer) | `count: 7`, JDK-internal frames |
| `threadId: 3, includeInternals: true` | `count: 7`, identical |

Thread 4 is empty with internals explicitly included, so JDI really does report zero frames for it (a native VM thread). And thread 3's JDK-internal frames come back even with `includeInternals: false`, so the filter is not involved either way. The empty result is truthful — it is only *unexplained*.

## Expected

Match the behavior the empty-stack handling already provides on the auto-anchored path: return the empty result **with a `note`** saying the requested thread reported no frames and suggesting a thread that does (the response already knows the full thread list). Something like:

> `note: "Thread 4 (Signal Dispatcher) reported no stack frames — it is a native VM thread. Threads with frames: 1 (main), 3 (Finalizer)."`

Echoing the requested thread id back in the response would help too: right now `lastStop.threadId` is the only thread id present and it refers to a different thread, which actively misleads.

## Impact

Small but squarely in the "agent can't tell what happened" category. An agent following the documented recovery — get a thread list, pick one, ask for its stack — can land on a frameless thread and receive a successful-looking empty answer with no hint that it should try a different id. That is the same failure shape the empty-stack `note` was introduced to eliminate; this path just wasn't covered.

Found during a `/testdebugger` sweep while exercising the AT9b `threadId` recovery.

debugmcpdev:
## Summary

- try JavaScript Local/Block, Closure, Module/Script, then Global scopes before walking down the stack
- filter rdbg `%` pseudo-variables by default so native pseudo-only frames do not block Ruby walk-down
- preserve explicit frameless-thread anchors while identifying the queried/adopted thread and suggesting a frame-bearing alternative
- document the returned `threadId` and fallback behavior

## Validation

- reproduced each regression twice before the fix
- `pnpm build`
- `pnpm test:unit` (4,333 passed)
- `pnpm lint`
- `pnpm changelog:check`
- `git diff --check`
- Docker JavaScript smoke (6 passed)

## Dogfooding

Debugged the compiled inspection policy through the source dev proxy and a local MCP SDK client, then exercised the real JavaScript and Ruby examples. This exposed js-debug's `Closure (name)` label variant, which is covered by the #548 regression because it is the same scope-fallback defect.

Closes #548
Closes #549
Closes #553


debugmcpdev:
## Review adjudication

**Fixed (blocker)**
- **JS Global fall-through.** Verified against a real js-debug trace of the #548 scenario: js-debug marks `Global` `expensive: true`, but the session layer has no expensive-skip and fetched its 139 variables (141 total for the frame, under the 300 budget). With Local/Closure/Module all empty the new ladder returned Node's globals as "locals" and — because the result was non-empty — the #468 frame walk-down never ran. Pre-PR, Global was reachable only for frames with no Local-ish scope at all (top-level script frames). That gate is restored: ladder is Local → Closure → Script/Module; Global is consulted only when the frame exposes no Local scope; otherwise `[]` so the walk-down proceeds.

**Fixed (smaller)**
- Ruby: filter exactly `%self` by default rather than every `%`-prefixed name — rdbg also exposes `%return` (after step-out) and `%raised` (exception stops), and hiding those loses the value the user stepped out to see (and would make a `%return`-only frame walk-down away).
- `canonicalScope` picked the first scope in the adapter's list matching *any* canonical name rather than the most-preferred one — latent (every Delve capture shows `Locals` alone) but wrong for any adapter listing e.g. `Closure (fn)` ahead of `Local`. Now ranked by `getLocalScopeName()` order.
- `describeFramelessThread` duplicated `scanThreadsForFrames` minus its `!thread` null guard; it now reuses the scan (one `threads` request) and the no-alternative wording no longer asserts "native/runtime thread" as the cause — the transient stack-not-materialized race gives the same empty answer.
- `threadId ?? currentThreadId` is **kept** (js-debug's real thread id is `0`; `||` silently redirected a targeted request) and now covered by a test.
- Stale JSDoc on `anchorNote` / `getStackTraceDetailed` and the `src/server.ts` "no scan" comment; `docs/tool-reference.md` `get_local_variables` updated for the JS ladder, Ruby `%self`, and `note`. Changelog fragment amended (it promised Global).

**Declined**
- Enforcing the no-re-anchor invariant inside `getStackTraceDetailed` — the implicit server path passes `currentThreadId` explicitly with `ensureStackReady`, so `typeof threadId` cannot distinguish the cases; the contract is documented instead.
- `Set`-identity `contributingScope` (same invariant the #438 truncation attribution relies on), merging multiple Closure scopes, a per-thread deadline on the alternative sweep (parity with the existing implicit-path scan), the double-scan perf nit, and `pause_execution`'s "0 = all threads" wording (that tool's own contract).
- `Block` / `Catch Block` / `With Block` exact-name matching — pre-existing `Block:` prefix on `main`; filed as #558.

PR #555, opened 2026-08-28
fix(go): normalize function breakpoint removal

### LINKED ISSUE #550: Go: remove_breakpoint {function: "main"} fails after set_breakpoint auto-qualified it to main.main (#467 follow-up)
## Summary

`#467` taught `set_breakpoint` to auto-qualify a bare Go function name (`main` → `main.main`) and to disclose the rewrite. `remove_breakpoint` was not given the same normalization, so it only matches the **stored qualified** name. An agent that sets a breakpoint with the name it typed cannot remove it with that same name.

## Repro

```
create_debug_session {language: "go"}

set_breakpoint {sessionId, function: "main"}
  -> success
     requestedName: "main"
     functionName:  "main.main"
     warning: "Auto-qualified function breakpoint 'main' to 'main.main' — Go resolves
               function breakpoints against package-qualified runtime names…"

start_debugging {scriptPath: ".../examples/go/hello_world.go",
                 dapLaunchArgs: {stopOnEntry: false, mode: "debug"}}
  -> stops inside main.main

remove_breakpoint {sessionId, function: "main"}
  -> ERROR: "No function breakpoint found for main"        <-- the bug

remove_breakpoint {sessionId, function: "main.main"}
  -> success: verified: true, adapterId: 1,
              boundFile: hello_world.go, boundLine: 11
```

`list_breakpoints` then correctly shows `functionBreakpoints: []`.

## Why this matters

The asymmetry is invisible to the caller. `set_breakpoint` accepted `main`, told the user it rewrote it, and stored it under the new name — but nothing about the accept/warn response suggests that the *removal* key has changed. Any agent doing the obvious set-then-remove round trip with the name it chose hits a hard error on a breakpoint that plainly exists.

It also breaks the symmetry the tool surface otherwise maintains: `remove_breakpoint` accepts the same addressing keys `set_breakpoint` does.

## Scope

Reproduced on both Go-capable group primaries in the 2026-08-28 testdebugger run:

| Server | Group/transport | F6 |
|---|---|---|
| mcp-debugger | local / http | FAIL |
| mcp-debugger-npx | npx / sse | FAIL |

(Docker image has Go disabled via `DEBUG_MCP_DISABLE_LANGUAGES`.)

## Suggested fix

Run `remove_breakpoint`'s `function` argument through the same `normalizeFunctionBreakpointName` policy hook `set_breakpoint` uses before matching, so a bare name resolves to the stored qualified record. Matching the qualified name directly should keep working. Worth applying the same normalization to any other lookup-by-function-name path.

## Everything else in #467's surface is intact

Both servers confirmed the rest of `#467`:

- F2 disclosed the auto-qualification (`requestedName` + `functionName` + `warning`).
- F8: a bare non-`main` name (`notARealFunc`) got the `#308` package-qualification hint on `set_breakpoint`, and `start_debugging` carried the "Function breakpoint(s) not bound at launch" warning naming it with its stored diagnostic; one server also surfaced it in `get_output`.

## Minor doc note (not part of this bug)

Delve anchors the `main` function breakpoint at `hello_world.go:11` (`func main() {`), not line 12 — the same function-entry-line behavior already documented for debugpy's `def` line. Both servers reported line 11.

Found by the `/testdebugger` suite, 2026-08-28. Related: #467, #308.

debugmcpdev:
## Summary

- normalize `remove_breakpoint.function` through the same adapter-policy hook used when setting function breakpoints
- preserve direct removal by qualified Go names
- disclose requested/effective names and normalization warnings, including useful no-match diagnostics

## Validation

- reproduced the regression twice before the fix
- `pnpm build`
- `pnpm test:unit` (4,328 passed)
- `pnpm lint`
- `pnpm changelog:check`
- `git diff --check`

## Dogfooding

Used the source dev proxy with the real Go example and Delve to set and remove a bare `main` function breakpoint, verify the disclosed `main.main` normalization, launch, pause, inspect frames, and continue.

Closes #550

### DIFF
--- a/changelog.d/550.fixed.md
+++ b/changelog.d/550.fixed.md
@@ -1,0 +1,1 @@
+**Go function breakpoints can be removed with the same bare name used to create them** — `remove_breakpoint {function: "main"}` now applies the same `main` to `main.main` normalization as `set_breakpoint`, discloses the effective name, and preserves direct qualified-name removal (#550)
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -279,11 +279,12 @@ Lists all breakpoints in a session with their current verified state and adapter
 
 ### remove_breakpoint
 
-Removes one breakpoint by id, or every breakpoint at a `file` + `line` location. Takes effect immediately while the program is running or paused (the file's remaining breakpoint set is re-sent to the adapter); also works after the program exits, so breakpoints can be adjusted before a relaunch.
+Removes one breakpoint by id, every function breakpoint with a given name, or every breakpoint at a `file` + `line` location. Takes effect immediately while the program is running or paused (the file's remaining breakpoint set is re-sent to the adapter); also works after the program exits, so breakpoints can be adjusted before a relaunch.
 
 **Parameters:**
 - `sessionId` (string, required): The ID of the debug session.
-- `breakpointId` (string, optional): Breakpoint id from `set_breakpoint` or `list_breakpoints`. Takes precedence over `file` + `line`.
+- `breakpointId` (string, optional): Breakpoint id from `set_breakpoint` or `list_breakpoints`. Takes precedence over `function` and `file` + `line`.
+- `function` (string, optional): Alternative addressing — remove **all** function breakpoints with this symbol name. The same per-language rewrite `set_breakpoint` applies is used here, so the name you set with is the name you remove with (a bare Go `main` matches the stored `main.main`); the response reports the effective name as `functionName` and, when it was rewritten, the original as `requestedName`.
 - `file` (string, optional): Alternative addressing — source file path (use together with `line`). Removes **all** breakpoints at that location.
 - `line` (number, optional): Alternative addressing — line number (use together with `file`).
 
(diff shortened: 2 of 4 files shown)


debugmcpdev:
## Review adjudication

**Fixed**
- Filter now matches the literal requested name **or** the rewritten one — a record stored un-rewritten (policy lookup failure, other entry path) was unreachable by name after this PR.
- Removal responses no longer carry the set-time "auto-qualified … never binds" note; disclosure is via fields, same shape as `set_breakpoint` (`functionName` always, `requestedName` only when rewritten).
- No-match branch keeps the removal loop's warnings and appends the per-adapter package-qualification hint (`getFunctionBreakpointNameHint`) when no rewrite applied — `remove_breakpoint {function:"compute"}` on Go now says to use `main.compute`.
- Tool schema and `docs/tool-reference.md` document `function` on `remove_breakpoint` (it was missing entirely) and the disclosure fields.
- Tests: set→remove round-trip through the real `GoAdapterPolicy`, bare-name stored record, no-match hint.

**Declined**
- Batching the per-record `syncFunctionBreakpoints` resend — pre-existing loop shape on `main`, not a regression; filed as #559.
- Moving name resolution into `SessionManager` — architecture preference, out of scope for a fix PR.

Issue #337, opened 2026-08-16
[BUG] Abandoned attach session can leave an orphaned lldb-server tracing the target (invisible to list_debug_sessions)


debugmcpdev:
Found while verifying the #332 ephemeral-sidecar flow on kind (cpp attach-by-PID inside a kubectl-debug ephemeral container).

## What happened

1. `attach_to_process {processId: 1, stopOnEntry: true}` succeeded; the MCP client then crashed without `detach`/`close_debug_session`.
2. A later `attach_to_process` on a fresh session failed with `Attach did not become debuggable: no threads reported within 5000ms (last failure: Proxy not initialized)`.
3. `list_debug_sessions` returned **no sessions** — nothing left to close via tools.
4. Yet the first session's proxy chain was still alive and still tracing the target: `ps` in the container showed the proxy worker + `codelldb` + `lldb-server`, and `grep TracerPid /proc/1/status` pointed at the lldb-server PID.
5. Recovery required killing the orphaned chain by PID (after which a fresh attach + function breakpoint worked normally).

## Why it's bad

The target process stays ptrace-claimed by a zombie debugger the tools can no longer see or clean up — every subsequent attach fails with a misleading "no threads reported" error. In a pod, the only in-band fix is restarting the pod.

## Notes / suspicions

- The exact trigger wasn't fully isolated: between the abandoned session and the failed re-attach there was also a second failed attach attempt. The observable contract violation is steps 3+4 together: a live, target-tracing proxy chain with zero sessions listed.
- Possibly `list_debug_sessions` hides ERROR/terminated sessions whose proxy is still running, or the session record is dropped on a failed attach-verify without tearing the proxy down.
- Worth checking: attach-verify failure path (`attachVerifyFailed`) and whatever removes the session record — does it await proxy shutdown?

Environment: mcp-debugger image (main @ #331/#328 branches) as a kubectl-debug ephemeral container, kind v0.32/containerd, cpp adapter, CodeLLDB 1.11.8.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #556, opened 2026-08-28
fix(attach): preserve proxy initialization diagnostics

### LINKED ISSUE #551: attach_to_process failures omit the #493 diagnostics (data.initProgress + data.proxyLogPath) that start_debugging returns
## Summary

`#493` added structured failure diagnostics to the tool result — `data.initProgress` (which init stage stalled) and `data.proxyLogPath` (where the full proxy log lives) — precisely because *"the agent reading the error has no other way to these facts"* (comment in `session-manager-operations.ts`).

That enrichment lives only in `startDebugging`'s catch block. `attachToProcess` fails through a different path and returns a bare result with **no `data` at all**, so an agent debugging a failed attach gets the prose message and nothing machine-readable — no stage facts, no log path.

## Repro

Start a dumb TCP listener that accepts the connection and then says nothing (the canonical proxy-init stall):

```bash
node -e 'require("net").createServer(()=>{}).listen(45999,"127.0.0.1")'
```

Then:

```
create_debug_session {language: "python"}
attach_to_process {sessionId, port: 45999}     # ~30s
```

Actual result (verified over raw JSON-RPC against `dist/index.js stdio`, so nothing is stripped by a client):

```json
{
  "success": false,
  "state": "error",
  "message": "Failed to attach: Debug proxy initialization did not complete within 30s. Connected to the debug adapter, but the \"initialize\" request never received a response. This is an adapter-side protocol stall, not a missing install — the adapter started and accepted the connection. Retrying usually succeeds; if it recurs, capture a DAP trace (DAP_TRACE=1) of the failing launch."
}
```

Expected: the same result plus

```json
"data": { "initProgress": { ... }, "proxyLogPath": "/tmp/debug-mcp-server/sessions/<id>/run-<ts>/proxy-<id>.log" }
```

## The message wording itself is correct

Worth stating explicitly, since it is the other half of `#493`: the prose **passes**. It names the stage actually reached, says the adapter connected and the `initialize` request went unanswered, and explicitly rules out a missing install. It does not carry a boilerplate attempt count. Only the structured payload is missing.

## Cause (verified in source)

`src/session/session-manager-operations.ts`:

- `startDebugging`'s catch (~L974–1090) reads `(error as {initProgress?}).initProgress`, computes `proxyLogPath` from `session.logDir`, assembles `diagnosticData`, and spreads it as `data` on the returned result.
- `attachToProcess` (~L2661+) has no equivalent. Its verification-failure path builds `reason` and does `throw new Error(reason)` (~L2921), and the proxy-init timeout thrown from `proxy-manager.ts` L387–392 — which *does* carry `error.initProgress` — is never unpacked. Nothing on the attach path ever sets `data`.

So both facts exist at the throw site; attach just drops them.

## Suggested fix

Factor the `catch` enrichment in `startDebugging` into a shared helper (`buildFailureDiagnostics(error, sessionId)`) and use it in `attachToProcess`'s failure path too. `proxy-manager.ts` already attaches `initProgress` to the error, and `session.logDir` is available on the attach path as well, so no new plumbing is needed.

Worth auditing the other tool paths that can surface a proxy-init timeout for the same omission.

## Note on the test suite

The `/testdebugger` N6 workflow asserts *"the failed result's `data` carries structured `initProgress` **and** `proxyLogPath`"* and drives it through `attach_to_process` — so N6 has been asserting a contract only the launch path satisfies.

Found by the `/testdebugger` suite, 2026-08-28. Related: #493, #517.

debugmcpdev:
## Summary

- factor proxy initialization diagnostics into a shared launch/attach helper
- capture attach `initProgress` and `proxyLogPath` before teardown
- preserve raw structured failure data through the MCP response without changing successful responses

## Validation

- reproduced the regression twice before the fix
- `pnpm build`
- `pnpm test:unit` (4,329 passed)
- `pnpm lint`
- `pnpm changelog:check`
- `git diff --check`

## Dogfooding

Ran a real MCP attach through the source dev proxy against the adversarial DAP scenario fixture, dropped the initialize response, and confirmed the failure retains transport progress and the proxy log path before cleanup.

Closes #551

### DIFF
--- a/changelog.d/551.fixed.md
+++ b/changelog.d/551.fixed.md
@@ -1,1 +1,1 @@
-Preserve proxy initialization progress and log-path diagnostics when attach requests fail.
+**Failed attaches report where proxy initialization stalled** — `attach_to_process` and `create_debug_session` with a `port` now return the same `data.initProgress` / `data.proxyLogPath` diagnostics `start_debugging` already gave on a proxy init failure, instead of a bare error message (#551)
--- a/changelog.d/551.fixed.md
+++ b/changelog.d/551.fixed.md
@@ -1,0 +1,1 @@
+Preserve proxy initialization progress and log-path diagnostics when attach requests fail.
--- a/src/server.ts
+++ b/src/server.ts
@@ -1352,13 +1352,21 @@ export class DebugMcpServer {
                     adapterConfig: args.adapterConfig,
                   });
 
+                  // Forward the attach payload the same way attach_to_process
+                  // does: structured failure diagnostics (initProgress /
+                  // proxyLogPath, issue #551) and the dropped-adapterConfig
+                  // warning (issue #450) must reach this entry point too.
+                  const attachData = attachResult.data as { warning?: string } | undefined;
+                  const attachWarning = attachResult.success ? attachData?.warning : undefined;
                   result = { content: [{ type: 'text', text: JSON.stringify({
                     success: attachResult.success,
                     sessionId: sessionInfo.id,
                     state: attachResult.state,
                     message: attachResult.success
                       ? `Created and attached ${sessionInfo.language} debug session: ${sessionInfo.name}`
-                      : `Created session but attach failed: ${attachResult.error || 'Unknown error'}`
+                      : `Created session but attach failed: ${attachResult.error || 'Unknown error'}`,
+                    ...(attachData ? { data: attachData } : {}),
+                    ...(attachWarning ? { warning: attachWarning } : {})
                   }) }] };
                 } catch (error) {
                   this.logger.error('session:attach-failed', {
--- a/src/session/session-manager-operations.ts
+++ b/src/session/session-manager-operations.ts
@@ -971,13 +971,14 @@ export abstract class SessionManagerOperations extends SessionManagerData {
         },
       };
     } catch (error) {
+      const diagnosticData = this.collectProxyFailureDiagnostics(session, error);
+      const initProgress = diagnosticData.initProgress as Record<string, unknown> | undefined;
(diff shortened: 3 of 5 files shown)


debugmcpdev:
## Review adjudication

**Fixed**
- `create_debug_session` with `port` (the inline attach path — the #551 repro shape) built its own response and dropped `attachResult.data`, so the new diagnostics never reached that caller. It now forwards `data` and surfaces `data.warning` at the top level like `attach_to_process`.
- Diagnostics are now collected **after** `stopProxyPreservingSession`, not before: teardown never clears `logDir` (the "can clear the per-run log directory" comment described an invariant that doesn't exist), and calling the helper first meant a throw inside it would skip the mandatory proxy teardown (#337). The unit test that manufactured that invariant by nulling `logDir` in its teardown mock is corrected.
- `collectProxyFailureDiagnostics` returns a typed `{ initProgress?: ProxyInitProgress; proxyLogPath?: string }` instead of `Record<string, unknown>`; the two casts in the launch catch are gone.
- Test fixtures used `pendingRequest`; the real `ProxyInitProgress` field is `pendingCommand`. Fixed, and the server-side test moved into its own `attach failure diagnostics (issue #551)` describe with a sibling for the `create_debug_session` pass-through.
- Changelog fragment brought to house style (bold lede + `(#551)`).

**Declined**
- Gating `proxyLogPath` on `pathExists` — the expected location is still the right pointer after a retry; the launch path already gates the *tail read* on existence.
- Attach-side `errorDetails` / proxy-log-tail logging parity with launch, and a shared `proxyLogPathFor` helper — worthwhile, but scope creep for this fix; filed as #561.

PR #557, opened 2026-08-28
fix(cpp): make forced source rebuilds safe on Windows

### LINKED ISSUE #552: cpp: adapterLaunchConfig.forceRebuild discards the adapter's LLDB launch transform and forwards the raw .cpp as program
## Summary

On a C/C++ **source-file launch**, passing `adapterLaunchConfig: { forceRebuild: true }` causes the cpp adapter's entire transformed LLDB launch configuration to be discarded. The raw `.cpp` path is forwarded to CodeLLDB as `program`, and the launch fails with an error naming a sibling `.exe` path the adapter never uses:

```
'C:\Users\johnf\projects\mcp-debugger\examples\cpp\hello_world.exe' does not exist
```

`forceRebuild` is the documented escape hatch for stale artifacts (CLAUDE.md: *"headers not staleness-tracked; `forceRebuild: true` to override"*), so the one knob provided for recovering from a stale build is the knob that breaks the launch.

## Reproduction

Host: Windows 11, MinGW-w64 g++ 15.2.0, local `dist/` build @ 002a49d2. Reproduced identically on the `npx` backend. One session at a time — **not** a concurrency artifact.

```jsonc
create_debug_session { "language": "cpp" }
start_debugging {
  "scriptPath": "<abs>/examples/cpp/hello_world.cpp",
  "dapLaunchArgs": { "stopOnEntry": false },
  "adapterLaunchConfig": { "forceRebuild": true }
}
```

→ `success: false`, `Proxy exited during initialization. Code: 0`

## Root cause, from the proxy log

The launch request that reaches CodeLLDB differs structurally between the two cases.

**Working** (`adapterLaunchConfig` omitted) — the adapter's full transformed config:

```json
{"args":[],"console":"internalConsole","cwd":".../examples/cpp","env":"<0 env vars redacted>",
 "initCommands":[],"justMyCode":true,"name":"Debug C/C++","noDebug":false,
 "postRunCommands":[],"preRunCommands":[],
 "program":"...\examples\cpp\.debug-mcp\hello_world.exe",
 "request":"launch","sourceLanguages":["cpp"],"sourceMap":{},
 "stopOnEntry":false,"terminal":"console","type":"lldb"}
```

**Broken** (`adapterLaunchConfig: { forceRebuild: true }`):

```json
{"args":[],"console":"internalConsole","cwd":".../examples/cpp","forceRebuild":true,
 "justMyCode":true,"noDebug":false,
 "program":".../examples/cpp/hello_world.cpp","stopOnEntry":false}
```

Everything the cpp adapter contributes is gone — `type: "lldb"`, `request`, `name`, `sourceLanguages`, `sourceMap`, `initCommands`/`preRunCommands`/`postRunCommands`, `terminal` — and `program` is still the source file. `forceRebuild` itself is forwarded verbatim to CodeLLDB as an unrecognized key rather than being consumed by the adapter.

The proxy log for the broken run contains **no compile activity at all** (no `g++`, no `.debug-mcp`): the source→binary step never runs, rather than running and failing.

CodeLLDB then derives a sibling binary from the `.cpp` and reports `'…\hello_world.exe' does not exist`. That path is doubly misleading — the adapter compiles into `examples/cpp/.debug-mcp/`, and no sibling `hello_world.exe` has ever existed:

```
$ ls examples/cpp/            # no hello_world.exe
hello_world.c  hello_world.cpp  .debug-mcp/
$ ls examples/cpp/.debug-mcp/
hello_world       # ELF   (from the Docker variants)
hello_world.exe   # PE32+ (from the Windows host)
```

## Trigger is specifically `forceRebuild`

| `adapterLaunchConfig` | result |
|---|---|
| omitted | PASS — `program` rewritten to `.debug-mcp\hello_world.exe` |
| `{}` | PASS |
| `{ "initCommands": [] }` | PASS |
| `{ "forceRebuild": true }` | **FAIL** |

So this is not "any `adapterLaunchConfig` replaces the config" — a non-empty config with a legitimate pass-through key merges correctly. Only `forceRebuild` derails it.

The same call **succeeds on all three Docker variants**, so the host and container paths diverge here.

## Expected

`forceRebuild: true` should force the source recompile and leave the rest of the adapter's launch transform intact — `program` still rewritten to the `.debug-mcp/` artifact. It should be consumed by the adapter, not forwarded to CodeLLDB.

Secondarily: when the compile step is skipped or fails, the error should name the artifact path the adapter actually intends to use, not a sibling path derived by CodeLLDB from the source name.

## Impact

The documented recovery for a stale/cross-platform artifact is unusable on the Windows host, and the error message sends you looking for a file in the wrong directory. The `/testdebugger` harness instructs passing `forceRebuild: true` for exactly this fixture, which is how this surfaced.

debugmcpdev:
## Summary

- propagate launch/attach transform failures instead of forwarding known-invalid generic configurations
- compile forced source rebuilds to unique staging executables before promotion
- retain and launch managed versioned artifacts when Windows locks the canonical executable, reuse the newest artifact, and clean older unlocked versions opportunistically
- preserve the previous executable after failed compilation and document the Windows fallback

## Validation

- reproduced both the transform fallback and artifact regressions twice before the fix
- `pnpm build`
- `pnpm test:unit` (4,331 passed)
- `pnpm lint`
- `pnpm changelog:check`
- `git diff --check`
- Docker C++ smoke (2 passed)

## Dogfooding

Kept the canonical Windows C++ executable running while forcing a source rebuild through the source dev proxy. The session launched the managed artifact, hit a breakpoint, exposed locals, stepped, and continued; a later launch reused that artifact. The Docker smoke also compiled the real C++ example and completed a breakpoint cycle.

Closes #552


debugmcpdev:
## Review adjudication

**Fixed (blockers)**
- **Adapter leak on transform throw.** `adapterRegistry.create()` runs before the transform; rethrowing exited `startProxyManager` before the adapter was handed to a `ProxyManager`, and nothing else ever calls `adapter.dispose()` — so every failed compile stranded a registry slot until "Maximum adapter instances (10) reached for language: cpp". This was a pre-existing leak class (the MSVC throw, `resolveExecutablePath`, and `buildAdapterCommand` share the window); it's now fixed once with a dispose-on-throw guard covering the whole window until ownership transfers.
- **MSVC structured path destroyed.** `evaluateToolchain` throws from *inside* `transformLaunchConfig` when `CPP_MSVC_BEHAVIOR`/`RUST_MSVC_BEHAVIOR=error`; the rethrow bypassed the `consumeLastToolchainValidation()` block, losing the `MSVC_TOOLCHAIN_DETECTED` sentinel, `canContinue`, `data.toolchainValidation`, the CREATED-state reset — and it skipped clearing a stale validation from a previous launch. The block is factored into `applyToolchainValidation()` and runs on both the success path and the catch.

**Fixed (smaller)**
- Removed the `typeof adapter.transformLaunchConfig === 'function'` else-branch: `IDebugAdapter.transformLaunchConfig` is required, and the branch only existed for three incomplete test doubles while re-introducing the untransformed-config forwarding this PR removes. Doubles fixed; the variable is now non-optional, which also retired the dead `??` fallback and `&&` guard.
- `promoteStagedOutput` unlinked the canonical exe *before* the rename on win32 — a rename failure after that destroyed the previous good binary. Now rename-first (`MoveFileEx(REPLACE_EXISTING)` replaces an unlocked target and fails on a locked one); the fallback warning carries the actual errno/message instead of a blanket "in use".
- `mkdir` of `.debug-mcp/` is guarded, so EACCES/EROFS returns through the normal `C/C++ compile failed: …` path instead of a raw errno.
- Changelog fragment brought to house style and now mentions the transform-failure behavior change.

**Declined**
- Concurrent-rebuild races / staging-file selectability / reuse-path cleanup of a "live" Linux artifact — pre-PR compiled straight to the canonical path, so two concurrent compiles were strictly worse, a crash-partial artifact was equally reusable, and managed artifacts are launched only when rename fails (Windows unlink of a running exe fails). Not regressions; the crash-partial / concurrent-cleanup case is filed as #560.
- `needsRecompile` bypassing the injected seam (tests use real temp dirs), `ArtifactFileSystem` vs `IFileSystem` (`IFileSystem` has no `rename`), double `readdir`.

Issue #678, opened 2026-09-09
js launch: skipFiles always blackboxes node_modules — a step from inside a dependency never lands and pause_execution on a launched HTTP server never stops (justMyCode: false is inert)


debugmcpdev:
Found dogfooding #672 on `fix/672-paused-frame-anchor` (main @ a27a6a50 + the #672 fix). JavaScript **launch** session of mcp-debugger's own HTTP server (`dist/index.js http -p 3111`), breakpoint inside a dependency at `node_modules/.pnpm/router@2.2.0/node_modules/router/index.js:160` (`const self = this` in `Router.prototype.handle`).

## Two symptoms, one mechanism

**1. A step from inside a dependency never lands.** Paused on the router breakpoint, `step_over` returns `pending: true` ("still executing after 5s"). The request completes (the client gets its response), the session stays `running`, and the *next* request stops at the breakpoint again with `stopReason: "breakpoint"` — the step is simply lost. Line 160 → 161 is a plain assignment; nothing long-running is being stepped over.

**2. `pause_execution` on a launch-mode HTTP server never lands.** With no breakpoints set, `pause_execution` on the idle server returns `pending: true` (expected). Then `curl http://127.0.0.1:3111/health` three times: every request is answered (`200`, or `403` for a disallowed Host — so node internals, express *and our own middleware* all executed) and the session never becomes `paused`. Two attempts, same result.

The per-session proxy log shows the mechanism. After the `pause` request:

```
paused-recorded=234   continued=234   stopped=0
```

i.e. 234 CDP `Debugger.paused` notifications (`[CdpFnBpBridge] Debugger.paused recorded`), each followed within ~1 ms by a DAP `continued` event, and not a single DAP `stopped`. js-debug pauses, sees the frame is in a skipped script, and resumes — over and over. The launch config it was given:

```
"skipFiles":["<node_internals>/**","**/node_modules/**"], "justMyCode":true, "smartStep":true
```

On an HTTP request path every frame is node internals or `node_modules` until user code is *called* (our middleware is called from express and returns to express). js-debug's skip handling only ever steps *out* of a skipped frame, so it never finds user code by returning into it; each stop lands in another skipped frame until the request is over and the event loop is idle. A `pause` and a `step` are both consumed that way.

## Why launch and not attach

[...]

Issue #346, opened 2026-08-16
[QUESTION] Go stack filtering can return an empty stack — should it keep the first frame like JS?


debugmcpdev:
Surfaced during the #344 review. `GoAdapterPolicy.filterStackFrames` (`packages/shared/src/interfaces/adapter-policy-go.ts`) is a plain `frames.filter()` with no fallback: a goroutine paused entirely inside `/runtime/` or `/testing/` frames yields an **empty** `get_stack_trace` result.

Policies currently disagree on the all-internal edge case:
- **JS**: returns `[frames[0]]` (never empty)
- **Java**: returns the full unfiltered array (never hides the stack)
- **Go, .NET**: may return `[]`

Docs now describe this accurately (fixed in #344), but the behavioral question stands: an empty stack is a dead end for an agent — it can't tell "no frames" from "all frames filtered", and `includeInternals: true` is only discoverable if it already knows the filtering exists. Options:

1. Give Go (and .NET) the JS-style first-frame fallback
2. Adopt Java's return-everything fallback everywhere
3. Keep as-is but annotate the empty result (e.g. a note that N internal frames were hidden and how to reveal them)

Option 3 fits the AX principle of replacing what the agent must know with what it can see in the response.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #679, opened 2026-09-09
fix(inspection): keep the paused frame when the display filter hides it; anchor locals/evaluate on it (#672)

### LINKED ISSUE #672: js: when the paused frame is inside node_modules, get_stack_trace hides it and get_local_variables/evaluate_expression silently anchor on an async ancestor frame
Found dogfooding #667 (mcp-debugger debugging its own HTTP server) on main @ e6bfebad, after #661.

## What happened

A breakpoint inside a dependency — the SDK's Host-validation middleware, `node_modules/@modelcontextprotocol/sdk/dist/esm/server/middleware/hostHeaderValidation.js:50` — fires (`stopped reason=breakpoint hitBreakpointIds=[1]` in the DAP trace). Then:

**`get_stack_trace`** (default `includeInternals: false`):
```
stackFrames[0] = { id: 66, name: "handleHttpCommand", file: "...\dist\cli\http-command.js", line: 299 }
... 4 more of our own frames
hiddenFrames: 60, note: "60 internal frame(s) hidden — pass includeInternals: true to see them."
stopReason: "breakpoint"
```
Nothing says the *paused* frame is one of the hidden ones. It reads as "paused on a breakpoint at http-command.js:299" — a line that has no breakpoint and is the `app.listen(...)` call, an async ancestor of the request.

**`get_local_variables`** (default): `variables: [], frame: handleHttpCommand:299, message: "The Local scope is empty."`. With `names: ["hostname","allowedHostnames","hostHeader"]`: all three in `notFound`. Silently answered about the wrong frame.

**`evaluate_expression`** (no `frameId`): `Invalid frame context: Unable to evaluate on async stack frame`.

**Workaround that works:** `get_stack_trace includeInternals: true` shows the real frame 0 — `id: 30, hostHeaderValidation.js:50` (at its pnpm realpath) — and `evaluate_expression { frameId: 30 }` returns `{"hostname":"mcp-debugger","allowedHostnames":["localhost","127.0.0.1","[::1]"],"hostHeader":"mcp-debugger"}`.

**Control:** paused at a breakpoint in our own code (`createHttpApp`), the same default `get_local_variables` call returns the frame's locals correctly. The failure is specific to the paused frame being classified internal.

## Why

#661 made every path with a `node_modules` segment internal for JS and dropped the JS first-frame fallback so the #346 `allFramesInternal` guarantee engages. That guarantee only covers the *all-internal* case. When the top frame is internal but user frames exist further down the (async) stack, the shared inspection anchor follows the *display* filter to the first visible frame — here an async-stack ancestor V8 cannot evaluate in — instead of the frame the debuggee is actually paused in. (`src/session/inspection/frame-anchor-resolver.ts`, `src/server/handlers/inspection-tools.ts`, `packages/shared/src/interfaces/adapter-policy-js.ts`.)

Debugging inside a dependency is a first-class use of this tool ("where IDEs can't go"); an agent doing it currently gets wrong data with no hint.

## Suggested fix

1. The inspection anchor (locals / scopes / evaluate without `frameId`) must be frame 0 of the *unfiltered* stack when the session is paused there. Display filtering must never move it.
2. `get_stack_trace` should keep the paused frame visible even when it is internal (a single dependency frame at the top costs nothing), or add an explicit `pausedFrame` / note: "paused inside a hidden frame `<path:line>`; pass `includeInternals: true` or `frameId: N`".
3. `get_local_variables` / `evaluate_expression` responses should say which frame they used and warn when it differs from the stopped location.

## Repro

```
create_debug_session {language:"javascript"}
set_breakpoint {file:"<repo>/node_modules/@modelcontextprotocol/sdk/dist/esm/server/middleware/hostHeaderValidation.js", statement:"if (!allowedHostnames.includes(hostname))"}
start_debugging {scriptPath:"<repo>/dist/index.js", args:["http","-p","3111"]}
continue_execution
curl -H "Host: mcp-debugger" http://127.0.0.1:3111/health     # pauses at the breakpoint
get_stack_trace / get_local_variables / evaluate_expression {expression:"hostname"}
```

debugmcpdev:
Fixes #672

## What

With `node_modules` frames classified internal (#661), a JavaScript breakpoint or step that landed *inside* a dependency of a running server was invisible: `get_stack_trace` showed only the async ancestors behind "N internal frame(s) hidden", and `get_local_variables` / `evaluate_expression` silently anchored on the first *visible* frame — an `app.listen` ancestor V8 cannot evaluate in — so locals came back empty and evaluate failed with "Unable to evaluate on async stack frame".

`FrameAnchorResolver` now distinguishes the mixed case (paused frame internal, ancestors visible) from the all-internal one (#346):

- **User-directed stop** (`breakpoint`, `function/data/instruction breakpoint`, `step`): the stopped thread's raw frame 0 is kept at the head of the filtered list, so every consumer of `frames[0]` — stack, locals, evaluate, the step/pause `location` — anchors where the debuggee actually stopped. The `note` says so; `hiddenFrames` counts only what stayed hidden.
- **Any other stop** (`pause`, `entry`, `exception`): the paused frame stays hidden — those routinely land in runtime frames (`runtime.gopark`, a JDK sleep, a panic unwinder) where the first user frame is the actionable one — but the `note` names the hidden frame and its `frameId`.
- Only the thread the stop event named qualifies; a sibling thread parked in runtime code is filtered exactly as before.

`StackTraceResult` gains an optional `pausedFrameInternal`. `EvaluateResult` gains `frame` (the frame the expression ran in, same shape as `get_local_variables`) on success **and** on failure, so "Unable to evaluate on async stack frame" now says which frame it tried.

## Verified live (mcp-debugger debugging its own `http` server)

Breakpoint at `node_modules/.pnpm/router@2.2.0/node_modules/router/index.js:160`, `curl /health`:

| | before (main) | after |
|---|---|---|
| `get_stack_trace` frame 0 | `handleHttpCommand` `dist/cli/http-command.js:353` (the `app.listen` async ancestor), note: "47 internal frame(s) hidden" | `handle` `router/index.js:160`, note: "Paused inside an internal frame … kept as frame 0 …; 46 internal frame(s) hidden" |
| `get_local_variables` | `[]`, "The Local scope is empty." | the router's 9 locals (`req`, `res`, `idx`, `next`, …), `frame` = router |
| `evaluate_expression req.method + ' ' + req.url` | `Invalid frame context: Unable to evaluate on async stack frame` | `'GET /health host=mcp-debugger'`, `frame` = router |

`step` inclusion was gated empirically: ruby, go, java and javascript e2e smokes green (22/22) with it in the rule.

Dogfooding this surfaced #678 (launch-mode `skipFiles` always blackboxes `node_modules`: a step from inside a dependency never lands and `pause_execution` on a launched HTTP server never stops).

## Tests

- `session-manager-dap.test.ts` → new `describe('Paused frame hidden by the display filter (issue #672)')`: breakpoint / step keep the frame; pause discloses; `includeInternals` bypass; all-internal still takes the #346 path; sibling thread untouched; locals anchor on the kept frame (and `names` finds them); locals after a pause come from the visible frame and say so.
- `session-manager-evaluate.test.ts`: evaluate anchors on the kept frame and returns `frame`; after a pause it stays on the visible frame and discloses.

https://claude.ai/code/session_01FTzA7RiaM7fS18WU8a8Pgj


debugmcpdev:
Follow-up commit addresses the `/code-review high` findings on 76f4a00e (8 confirmed, 3 plausible, 2 cleanups):

| finding | outcome |
|---|---|
| `debugger;` arrives as `pause` → hidden branch | **fixed** — `JsDebugAdapterPolicy.normalizeStopReason` maps `pause` + "Paused on debugger statement" to `breakpoint` (rawReason kept); also stops it being auto-continued as an incidental first stop |
| stop-reason gate leaves a js `exception` inside a dependency broken | **fixed** — new `AdapterPolicy.isAsyncBoundaryFrame` (js: the sourceless line-0 separators); a boundary between the paused frame and the first visible one keeps the frame whatever the reason. Go/Java runtime landings with a synchronous caller visible keep today's anchor |
| `isStoppedThread` fallback keyed on caller-passed threadId | **fixed** — compares against the thread current at the start of the resolve; test with a threadId-less stop asserts explicit and implicit resolves agree |
| note advertises `includeInternals`/`frameId` to tools that lack them; duplicated sentence on get_stack_trace | **fixed** — resolver emits `pausedFrame` + neutral `pausedFrameNote`; each tool appends its own advice; handler owns the includeInternals sentence (asserted once) |
| kept frame with empty locals → contradictory walk-down note | **fixed** — one sentence naming both anchors; inverse-fixture test added |
| `frame`/`anchorNote` missing on two failure returns | **fixed** — single `anchorFields` spread on every default-anchored return; tests for the evaluate-throws, no-body and frameless cases |
| `lastStop` read after the await | **fixed** — snapshot after the PAUSED gate; a stop landing mid-request falls back to plain filtering (test) |
| Java step_into keeps a JDK frame with a relative label | **fixed** — kept frame with a non-absolute, non-placeholder `file` gets `unresolvedSource: true` (test); the bridge's missing step class-exclusion is its own ticket |
| tool-schema descriptions not updated | **fixed** — three descriptions, 24 snapshots regenerated |
| docs: 47 vs 46, step list, `anchorNote` wording | **fixed** |
| two stop-reason taxonomies; four `{name,file,line}` literals | **fixed** — `BREAKPOINT_STOP_REASONS` in `@debugmcp/shared`; `FrameSummary` |
| pause_execution/step `location` on the hidden branch has no note | **documented** (tool-reference caveat, changelog reworded) rather than widening `PauseResultData` in this PR |

Issue #673, opened 2026-09-09
js launch: breakpoints that demonstrably fired stay "Unbound breakpoint" — source-mapped bp is verified under its generated path, node_modules bp never gets a breakpoint event


debugmcpdev:
Found dogfooding #667 on main @ e6bfebad (after #495/#500). JavaScript **launch** session, `DAP_TRACE=1`.

## What happened

Two breakpoints, both of which demonstrably fired (the session paused on each, `stopped hitBreakpointIds=[2]` then `[1]`), are reported by `list_breakpoints` as `verified: false, message: "Unbound breakpoint"` — before launch, at the first pause, at the second pause, and after both had fired:

| breakpoint | requested | fired? | list_breakpoints |
|---|---|---|---|
| 1 | `node_modules/@modelcontextprotocol/sdk/dist/esm/server/middleware/hostHeaderValidation.js:50` (pnpm symlink path) | yes | Unbound |
| 2 | `src/cli/http-command.ts:349` (source-mapped; `dist/cli/http-command.js.map` exists) | yes, at `dist/cli/http-command.js:277` | Unbound |

`verified` is the only binding feedback an agent gets; here both signals are inverted.

## What the DAP trace shows

```
18:44:45.370 [parent out]  setBreakpoints hostHeaderValidation.js [50]
18:44:45.371 [parent in]   → [(id 1, verified false, 'breakpoint.provisionalBreakpoint')]
18:44:45.372 [parent out]  setBreakpoints src/cli/http-command.ts [349]
18:44:45.373 [parent in]   → [(id 2, verified false, 'breakpoint.provisionalBreakpoint')]
18:44:45.792 [child out]   setBreakpoints hostHeaderValidation.js [50]   → same provisional stub (id 1)
18:44:45.793 [child out]   setBreakpoints http-command.ts [349]         → same provisional stub (id 2)
18:44:45.834 [child out]   setBreakpoints (re-send) hostHeaderValidation.js [50] → []   (js-debug's empty no-change echo)
18:44:45.835 [child out]   setBreakpoints (re-send) http-command.ts [349]        → []
18:44:46.324 [child in]    EVENT breakpoint reason=changed id=2 verified=true line=277 path=c:\...\dist\cli\http-command.js
18:44:46.458 [child in]    EVENT stopped reason=breakpoint hitBreakpointIds=[2]
[...]

Issue #655, opened 2026-09-02
js get_stack_trace: includeInternals:false still returns ~50 node_modules frames, and source-mapped frames show unopenable relative '../src/…' paths


debugmcpdev:
Found while dogfooding: a `javascript` attach session on a live mcp-debugger HTTP backend, paused on a breakpoint in `dist/server/tool-result.js` (inspecting #647).

## What happened

`get_stack_trace` (default `includeInternals: false`) returned **62 frames** with `hiddenFrames: 22`. Of the 62 shown, only the top frame was the user's code; the rest were:

- ~40 dependency frames from `node_modules/.pnpm/router@2.2.0/…`, `express@5.2.1/…`, `body-parser@2.2.2/…`, `raw-body@3.0.2/…`, `@hono/node-server@2.0.12/…` (`handleRequest`, `next`, `dispatch`, `processParams`, `trimPrefix`, `jsonParser`, …)
- ~10 frames whose `file` is a **relative, non-existent path**: `../src/shared/protocol.ts`, `../src/server/webStandardStreamableHttp.ts`, `../src/server/streamableHttp.ts`, `../src/server/index.ts` (the MCP SDK ships `.js.map` files whose `sources` point at `.ts` files that are not in the package), and even mcp-debugger's own `../src/server/handlers/execution-tools.ts`, `../src/server/tool-dispatch.ts`, `../src/cli/http-command.ts` — relative to nothing an agent can open.
- a handful of `<unknown_source>` `await` / `Promise.then` / `bound-anonymous-fn` frames at line 0.

The agent-facing result is a 6 KB stack of which one line matters, plus file paths that fail if passed to `get_source_context` or a Read tool.

## Why

`JsDebugAdapterPolicy.isInternalFrame` (`packages/shared/src/interfaces/adapter-policy-js.ts:160-163`) recognizes only `<node_internals>`:

```ts
const filePath = frame.source?.path || '';
return filePath.includes('<node_internals>');
```

`node_modules` frames and `<unknown_source>` async-glue frames are therefore "user" frames. (The Go and Java policies already classify runtime/library frames more broadly — #369 did the same for LLDB/libc on rust/cpp.)

For the paths: js-debug source-mapped the frames and reported the map's raw `sources` entry as `source.path` when the mapped file could not be resolved. The launch transform sets `resolveSourceMapLocations: ['**', '!**/node_modules/**']`; the attach path appears not to, so dependency maps get applied.

## Suggested fix

[...]

Issue #574, opened 2026-08-29
pause(): pausePending is never cleared when the stop lands during thread discovery; proxyManager re-read after an await without a null check


debugmcpdev:
## Summary

Two pre-existing edges in `pause()` (now `src/session/execution/execution-controller.ts`, moved verbatim by #567):

1. `session.pausePending = true` is set unconditionally AFTER the awaited `threads` discovery request. If the stop event already landed during that await, the race guard `if (session.state === PAUSED && !stopEventSeen)` settles the call successfully but `pausePending` is never cleared, so the next stop is misattributed as the pending pause.
2. `const proxyManager = session.proxyManager` is re-read after the same await with no null check, but the core clears `session.proxyManager = undefined` on close/terminate/exit — a session that terminates during discovery throws a TypeError out of `pause()` instead of returning the usual "no active proxy" result.

## Suggested fix

Set `pausePending` before issuing the discovery request and clear it in every settle path; re-check `session.proxyManager` after the await and route through the existing `ProxyNotRunningError` path.

Found while adjudicating the /code-review findings on #567.

PR #686, opened 2026-09-10
fix(javascript): launch skipFiles honour justMyCode and caller lists, sourceMaps on by default, skipped-frame hints (#678, #684)

### LINKED ISSUE #678: js launch: skipFiles always blackboxes node_modules — a step from inside a dependency never lands and pause_execution on a launched HTTP server never stops (justMyCode: false is inert)

### LINKED ISSUE #684: js launch: sourceMaps defaults to false for .js programs — a compiled TypeScript app stops in dist/*.js frames while its src/*.ts breakpoints bind; js-debug's own default is true
Found dogfooding #671/#673 on `fix/671-sse-host-allowlist` (main @ a27a6a50 + the SSE fix). JavaScript **launch** of a compiled TypeScript program: `dist/index.js` with `dist/**/*.js.map` next to it and `src/**/*.ts` present.

## What happens

`packages/adapter-javascript/src/javascript-debug-adapter.ts:460-475` enables source maps only when the *program* is a TypeScript file:

```ts
if (isTS) {
  result.sourceMaps = true; ...
} else {
  const sm = Boolean(u.sourceMaps as unknown);   // undefined → false
  result.sourceMaps = sm;
  ...
}
```

so a `.js` entry point runs js-debug with `sourceMaps: false` unless the caller passes `adapterLaunchConfig: { sourceMaps: true }`. js-debug's own default (and VS Code's) is `true`. Two consequences, measured on the same target (`dist/index.js sse -p 3113`, breakpoint set by statement on `src/cli/host-allowlist.ts:218`, `curl -H "Host: evil.example"`):

| | default (`sourceMaps` unset → `false`) | `adapterLaunchConfig: { sourceMaps: true }` |
|---|---|---|
| does the `.ts` breakpoint fire? | yes — js-debug still binds it (through its breakpoint predictor) | yes |
| `get_stack_trace` frame 0 | `dist/cli/host-allowlist.js:159` `<anonymous>` | `src/cli/host-allowlist.ts:218` |
| the rest of the stack | `dist/cli/sse-command.js:189`, `dist/index.js:106`, … | `src/cli/sse-command.ts:265`, `src/index.ts:131`, … |
| `list_breakpoints` while paused on it | `verified: false, "Unbound breakpoint"` (the #673 symptom: js-debug's verification arrives under the generated path and matches nothing) | `verified: true, adapterId: 1` immediately |
| `get_local_variables` / step locations | generated file and lines | source file and lines |

So an agent sets breakpoints in one view of the program (`src/*.ts`, the only view `set_breakpoint`'s `statement:` addressing can use) and gets every stop, frame, local and step location reported in another (`dist/*.js`). Nothing in the responses says the two are the same code. #673's source-mapped half is a direct consequence: with source maps on, js-debug's `breakpoint` event carries the `.ts` path and the store matches it by path+line, no provisional-id correlation needed (the fix in #681 still matters for the `node_modules` no-event case and for anyone who launches with maps off).

## Suggested fix

- Default `sourceMaps` to `true` for `.js` programs too — js-debug's default — with `outFiles` defaulting the way the `isTS` branch already does (`determineOutFiles`, excluding `node_modules`) and `resolveSourceMapLocations: ['**', '!**/node_modules/**']`. A program without maps is unaffected (js-debug finds none). Maps whose sources are not on disk already fall under the #655 `unresolvedSource` handling.
- Keep `sourceMaps: false` as an explicit opt-out for launches that want generated locations.
- Docs (`docs/javascript/README.md`, tool-reference `start_debugging`): say which view stops are reported in and how to switch.

## Repro

```
create_debug_session {language:"javascript"}
set_breakpoint {file:"<repo>/src/cli/host-allowlist.ts", statement:"const message = hostRejectionMessage(hostname, allowed);"}
start_debugging {scriptPath:"<repo>/dist/index.js", args:["sse","-p","3113"]}          # frames in dist/*.js, breakpoint "Unbound"
# vs
start_debugging {scriptPath:"<repo>/dist/index.js", args:["sse","-p","3113"], adapterLaunchConfig:{sourceMaps:true}}   # frames in src/*.ts, breakpoint verified
curl -H "Host: evil.example" http://127.0.0.1:3113/health; get_stack_trace; list_breakpoints
```

debugmcpdev:
## Summary

Two dogfood tickets filed while debugging mcp-debugger's own `dist/index.js http` under the JavaScript adapter, fixed together because they share one function (`transformLaunchConfig`), one test file and one doc page.

- **#678** — the launch transform always merged `**/node_modules/**` into js-debug's `skipFiles`, so `dapLaunchArgs.justMyCode: false` was inert: a `step_over` issued at a breakpoint inside a dependency never landed, `pause_execution` on a launched HTTP server never stopped, and the `pending: true` responses could not tell a slow step from a lost one.
- **#684** — launch `sourceMaps` defaulted to `false` for `.js` programs (js-debug's own default is `true`), so a compiled-TypeScript app reported every stop, frame and local in `dist/*.js`.

### The mechanism (settled by the review's live runs)

js-debug's resume-on-skipped-frame lives inside its **smart-stepper**: `getSmartStepDirection` returns early unless `launchConfig.smartStep`, and only then reads whether the frame is blackboxed. Node internals are on every default skip list, an idle server always pauses in internals, and a request path enters user code by calls, not returns — so with the stepper on, neither a pause nor a step issued from a skipped frame lands. `skipFiles` decides *what* is skipped; `smartStep` decides whether js-debug steps past it. `justMyCode: false` now turns both knobs (the pair attach got in #513). Whether the launch default should flip to `smartStep: false` like attach is #687.

### What changed

| Area | Change |
|---|---|
| `packages/shared/src/interfaces/js-launch-defaults.ts` (new) | `resolveJsLaunchSkipFiles` (a caller `skipFiles` **replaces** the default list, VS Code launch.json semantics; default is `<node_internals>/**` plus `**/node_modules/**` unless `justMyCode === false`), `resolveJsLaunchSmartStep` (explicit boolean wins, else follows `justMyCode`), `jsLaunchBlackboxesNodeModules`, `jsLaunchSkipsNodeInternals`. Single source of truth for the transform, the policy's launch flow and the hints. |
| `javascript-debug-adapter.ts` | `skipFiles` and `smartStep` from the helpers; `sourceMaps` defaults to `true` for every launch with js-debug's default `outFiles` and `resolveSourceMapLocations`; `sourceMaps: false` opts out (now honoured for `.ts` too) and forwards a caller `outFiles` untouched. |
| `AdapterPolicy.describePendingStop` (new optional hook) | The js policy explains a pending pause or step **only while the effective `smartStep` is on and the frame in question is actually skipped**: a pause on a launch that skips internals; a step from a skipped internals frame (remedy: `continue_execution` with a breakpoint in your code, or `adapterLaunchConfig.smartStep: false`); a step from a blackboxed node_modules frame (remedy: `justMyCode: false`). Silent once the stepper is off, for attach, and for a caller's own globs (no glob matcher in shared; an over-match would be a false hint). The controller appends the text to the grace-window message; the origin frame is read only when the hook exists, bounded to a fifth of the grace window, so other adapters pay nothing and their messages are byte-identical. |
| Step result disclosure | A step whose recorded stop is a breakpoint or exception is reported as `Stepped over; stopped on 'breakpoint' rather than on the step itself (see stopReason).` with `stopReason`/`rawStopReason`, plus `The program is back at the line the step was issued from.` when origin and landing match — instead of a bare `Stepped over` at a line the step never reached (found during the baseline). The wording is neutral because js-debug and debugpy both report `breakpoint` when a step lands on a breakpointed line (verified live). The controller owns the step wording now (`STEP_KINDS.successMessage` is the user-facing text; the handler passes it through). |
| Step preamble | Re-checks terminated / proxy / PAUSED after the origin read, as pause() does after its threads request (#574), so a concurrent `continue_execution` yields `Not paused` instead of a step sent to a running thread. |
| Schema / docs / changelog | `dapLaunchArgs.justMyCode` description (24 tool-list snapshots, description-only); `docs/javascript/README.md` launch table for `justMyCode` / `skipFiles` / `smartStep` and a rewritten TypeScript Support section; `docs/tool-reference.md` (justMyCode bullet, step response shape, pending-step and pending-pause hints, attach `justMyCode` inert on js); `docs/architecture/api-reference.md` `StepResultData`; three `changelog.d` fragments. |
| Tests | Unit: helper truth tables, transform cases, 12 policy hint cases, controller/handler disclosure cases (origin-line and other-line), preamble re-check and bounded origin read, defaults layer. E2E on a new `examples/javascript/express_idle_server.js` fixture: lost step explained (default), step lands (`justMyCode: false`), pause lands (`justMyCode: false`). |

### Measurements (mcp-debugger debugging its own `dist/index.js`)

| Probe | Before (main 25a54f5f) | After |
|---|---|---|
| `step_over` from a breakpoint in `router/index.js:160`, default launch | `pending: true`, generic "still executing"; step lost, request completed | `pending: true` + "The step was issued from a skipped frame (…router/index.js:160); this launch blackboxes node_modules (justMyCode: true) and js-debug's smart-stepper keeps stepping while the program is in skipped code, so on a request path the step may never land. Relaunch with dapLaunchArgs.justMyCode: false …" |
| same, `justMyCode: false` | identical to default (flag inert) | **lands** on `router/index.js:161`, `stopReason: step`; frames below read `src/cli/http-command.ts:454`, `src/index.ts:137` |
| `pause_execution` on the idle launched server, default | `pending: true`, generic; 164 paused/164 resumed in js-debug telemetry, never stopped | `pending: true` + "js-debug's smart-stepper is on for this launch (smartStep: true …) and Node internals are skipped, so a pause that lands in internals … may never land. … relaunch with dapLaunchArgs.justMyCode: false (or adapterLaunchConfig.smartStep: false) …" |
| same, `justMyCode: false` **without** the smartStep coupling | — | still churns (555 paused-recorded, 1110 continued): the pause lands in node internals and the stepper steps past it |
| same, `justMyCode: false` **with** `smartStep: false` derived | — | **lands** within the grace window: `stopReason: pause`, `<node_internals>/internal/async_hooks:230` (the truthful internal-frame pause attach got in #513) |
| review run A: default skip list + `smartStep: false` | — | pause **lands** (`<node_internals>/internal/timers:527`) — the stepper is the gate |
| review run B: internals-only skip list + `smartStep: true` | — | the chase reproduces with node_modules **not** blackboxed (323 paused / 646 continued / 0 stopped) |
| `.ts` breakpoint on `src/cli/host-allowlist.ts`, `sse` server, evil Host header | frame 0 `dist/cli/host-allowlist.js:159`, bound via `boundFile`, locals/step in `dist` | frame 0 `src/cli/host-allowlist.ts:218`, verified under its own path; locals, `evaluate_expression`, `step_over` → `src/cli/host-allowlist.ts:219` |
| breakpoint placed in the generated `dist/cli/host-allowlist.js` with maps on | — | binds and fires; frame 0 stays at the generated location, frames 1..n map to `src/*.ts`; store matching (#673/#681) unaffected — documented |
| step 10→11 in `simple_test.js` with a breakpoint on 11 | `Stepped over` | `Stepped over; stopped on 'breakpoint' rather than on the step itself (see stopReason).` — no false "cut short" claim |
| lost step, next request re-hits `express/lib/application.js:177` within the grace window | `Stepped over` at the same line, no disclosure | same wording + `The program is back at the line the step was issued from.` |

### Decisions

- **`skipFiles` replace semantics** (behaviour change, `changelog.d/678-skipfiles-replace.changed.md`): callers who passed `skipFiles` used to get theirs + both defaults; now exactly theirs. Include `<node_internals>/**` yourself to keep internals skipped.
- **`smartStep` coupled to `justMyCode`** unless set explicitly: `skipFiles` alone did not let a pause land (row 4). Cost: with the stepper off, steps also stop in unmapped generated helpers it used to skip — documented in the README row and the fragment.
- **Launch default stays `smartStep: true`** in this PR; the flip attach got in #513 is #687 (product decision with an unmeasured cost, now that `sourceMaps` is on by default).
- **`sourceMaps: true` for every program**: js-debug's own default; a program without maps is unaffected; maps whose sources are missing already carry `unresolvedSource` (#655).
- **Hints via one optional policy hook**, mirroring `isAsyncBoundaryFrame` (#672); the exact-match `stepStillRunning(5)` / `pausePending(5)` tests for other adapters keep passing.
- **Not done here**: `doctor` reporting anything about this; the generated-file frame-0 behaviour above is js-debug's own and is covered by one doc sentence.

### Review

`/code-review high` produced 10 findings (all real); 9 fixed in the follow-up commit, 1 deferred to #687. Row-by-row adjudication in the PR comment below.

### Gates

`npm run build && npm run typecheck:all && npm run lint && npm run check:docs && pnpm changelog:check` → exit 0 (ratchet unchanged). `npm run test:unit` → 263 files pass. `tests/e2e/mcp-server-smoke-javascript.test.ts` (all, incl. the three new #678 tests) and `tests/e2e/mcp-server-smoke-python.test.ts` (non-JS step path) pass locally; CI runs no e2e suites (the coverage script excludes them).

Closes #678
Closes #684

https://claude.ai/code/session_01FTzA7RiaM7fS18WU8a8Pgj


debugmcpdev:
## Review adjudication (`/code-review high`: 8 finder angles → 15 candidates → 8 verifiers, one of which ran three live sessions against this build → 10 findings)

All ten findings were real. Nine are fixed in the follow-up commit; one is deferred to #687 with the verifier's measurements. The four low-severity candidates the review cut for its cap are also handled where cheap.

| # | Finding | Verdict | What changed |
|---|---|---|---|
| 1 | `step()` awaits a real DAP `stackTrace` (origin read) between the pre-flight checks and the send, with no re-check; a concurrent `continue_execution` could drive a healthy session to ERROR; a wedged read could hold the step for the 30s DAP timeout | **Fixed** | Re-check after the awaits in pause()'s order (#574): terminated → `SessionTerminatedError`, proxy gone → `ProxyNotRunningError`, not PAUSED → `Not paused`, nothing sent. The origin read is bounded to a fifth of the step grace window (1s at the default) and the step proceeds without it. Two coverage tests. |
| 2 | The `ownStop` branch fires for every adapter whenever the recorded reason is not `step`, so the routine "step onto a breakpointed line" got the lost-step wording | **Fixed (wording)** | The wording is neutral and true in every case: `Stepped over; stopped on 'breakpoint' rather than on the step itself (see stopReason).`, plus `The program is back at the line the step was issued from.` only when origin and landing match (js, where the origin frame is already read). No claim that the step was cut short; `stopReason` is always forwarded. |
| 3 | Launch keeps `smartStep: true` while attach defaults it false, and js-debug's resume-on-skipped-frame is gated on `smartStep`, not `skipFiles` | **Deferred → #687** | Flipping the launch default is the product decision #678 left open and this PR took the other branch; the cost the verifier confirmed (stops in unmapped tsc helpers, more common with maps on) is unmeasured and `step_into` under `smartStep: false` is unverified. #687 carries runs A/B/C and a measurement plan. This PR fixes the internal inconsistency: hints, docs, schema and fragment all name the smart-stepper as the mechanism now. |
| 4 | `describePendingStop` keyed both hints on node_modules blackboxing; wrong in both directions (fired with `smartStep: false`, silent with `justMyCode: false` + `smartStep: true`); the internals branch fired when internals were not skipped | **Fixed** | New shared `resolveJsLaunchSmartStep` and `jsLaunchSkipsNodeInternals`, used by the transform, the hook, and the policy's third `smartStep` default. Hints only while the effective `smartStep` is on and the frame in question is actually skipped. A caller's own glob (`**/vendor/**`) still gets silence rather than a guess: this package has no glob matcher and an over-match would be a false hint. 12 policy tests. |
| 5 | Pause hint asserted "does not land at all" with no evidence gating, contradicting `pausePending`'s "will report paused once the stop lands" | **Fixed** | "may never land", and only under the smartStep + internals-skipped condition from finding 4. A program blocked in native code gets the base message alone once the stepper is off. |
| 6 | `stepStoppedOn` claimed "the location is where the program stopped", false for an exception stop in a hidden frame (the display filter's ancestor) | **Fixed (claim removed)** | The new wording makes no claim about `location`; it keeps its #672 display semantics (first visible frame), documented in the tool reference. |
| 7 | Internals step hint said "Use step_out", circular for a pending `step_out`; "only steps out" was inexact (js-debug repeats the issued direction, step-out after 256 tries) | **Fixed** | Remedy is `continue_execution` with a breakpoint in your own code, or `adapterLaunchConfig.smartStep: false`; mechanism text says the stepper keeps stepping while the program is in skipped code. |
| 8 | Two layers owned the step wording through the same predicate (controller picks the message, handler re-runs `isStepInterruptingReason` to decide whether to trust it) | **Fixed** | One owner: `STEP_KINDS.successMessage` is the user-facing `Stepped over/into/out`, the handler passes `data.message` through whenever present and falls back to its default only without one. `isStepInterruptingReason` is gone. |
| 9 | `isStepInterruptingReason` duplicated `USER_BREAK_REASONS` 70 lines up | **Fixed** | `USER_BREAK_REASONS` is exported and is the only definition. |
| 10 | Docs drift: step tools' response shape, "Pending steps", pause pending hint, `api-reference.md` `StepResultData` | **Fixed** | All four documented, plus the attach section now says `justMyCode` is inert on a js attach (the cut candidate), the `justMyCode` bullet and schema description name the mechanism, and the 684 fragment records that an explicit `sourceMaps: false` is honoured for `.ts` now. |
| cut | `defaultDapLaunchArgs` layer missing from the hint's merge (embedder-only) | **Fixed** | `ExecutionContext` exposes the defaults; the hook sees `{...defaults, ...dapLaunchArgs, ...adapterLaunchConfig}` like the launcher. |
| cut | `isSameLine` folded case on every platform; `samePath` already existed | **Fixed** | Uses `samePath` from the breakpoint layer. |
| cut | `outFiles` glob walk | **Refuted by the verifier** | The vendored bundle empties the pattern list when `rootPath` is undefined in DAP-server mode; nothing to do. |

Verification after the fixes: build, typecheck ratchet (unchanged), lint, `check:docs`, `changelog:check`, full unit suite, and both smoke suites (JavaScript incl. the three #678 e2e tests, Python for the non-JS step path) — results in the PR body. Note: CI runs no e2e suites (the coverage script excludes them), so the smoke suites are local evidence only.

PR #698, opened 2026-09-11
fix(cli): validate --port and announce the port the socket bound (#689)

### LINKED ISSUE #689: http/sse: --port is not validated (-p abc reports Node's options.port … NaN; -p 3001x silently binds 3001; -p 0 logs port 0)
Found while dogfooding #680 / #688 (the `--bind` work). `http` and `sse` take `-p, --port <number>` with a string default and no parser (`src/cli/setup.ts`), and both handlers do `parseInt(options.port, 10)` and hand the result straight to `listen()`.

## Measured (`dist/index.js` at 35892e23, Windows)

| Command | What the operator sees | Problem |
|---|---|---|
| `http -p abc` | `mcp-debugger: Failed to start server in HTTP mode: options.port should be >= 0 and < 65536. Received type number (NaN).` | Node's `RangeError` text: names `options.port` (an internal), reports `NaN` instead of the value typed, never mentions `--port` / `-p`. Same for `sse -p abc`. |
| `http -p 70000`, `http -p -5` | same line with `70000` / `-5` | The value is right, the knob is not named; the generic "Failed to start" path, not a named refusal like `--bind` / `--allowed-host` get (`BindAddressError` / `AllowedHostError` → "`<knob> value '…' …`. The server was not started."). |
| `http -p 3001x` | binds 3001 silently (here: `Port 3001 is already in use on 127.0.0.1`) | `parseInt` truncates trailing junk; a typo in a port is accepted as a different port. |
| `http -p 0` | log: `Debug MCP Server (HTTP) listening on 127.0.0.1:0`; `/health` → `listening: { port: 52914 }` | The startup line prints the requested port; the OS-assigned one is visible only in `/health` (which #688 made report the bound socket). |

## Suggested fix

- A shared `portOption()` beside `bindOption()` in `src/cli/`, with an `argParser` that accepts only `^\d+$` in `0..65535` and throws a commander `InvalidArgumentError` naming the flag and the typed value (`--port value 'abc' is not a port number (0-65535)`), so both commands and the help text stay in step. Commander prints `error: option '-p, --port <number>' argument 'abc' is invalid. …` and exits 1 before either handler runs — no server built, nothing to tear down.
- The `listening on …` line could read the port from `server.address()` in the `'listening'` handler that #688 added (the `listen` callback runs synchronously in the unit-test mocks, which is why the handler was added on the server instead).

## Not a regression

Pre-existing: the option has had no parser since the transport commands were added; #688 only made the `-p 0` case observable via `/health.listening`.

debugmcpdev:
Closes #689.

## What changed

`http` and `sse` took `-p, --port <number>` with no parser and handed `parseInt(options.port, 10)` to `listen()`. Measured on `main` (2a43e0a6):

| Command | Before | After |
|---|---|---|
| `http -p abc` | `Failed to start server in HTTP mode: options.port should be >= 0 and < 65536. Received type number (NaN).` | `--port value 'abc' is not a port number; give an integer 0-65535 (0 lets the OS choose). The server was not started.` (exit 1) |
| `http -p 3001x` | silently bound **3001** | refused, same line, nothing bound |
| `http -p 70000`, `-p -5` | Node's text, knob not named | refused by name |
| `http -p 0` | log: `listening on 127.0.0.1:0`; the OS port visible only in `/health` | log: `listening on 127.0.0.1:52423`, `endpoint available at http://127.0.0.1:52423/mcp`; `/health.listening.port` = 52423 |

Same for `sse`.

**Design.** The issue sketched a commander `argParser`. This PR follows the repo's own idiom for a bad CLI value instead — a resolver that throws a named error, caught by the handler, which logs it, `reportFatal`s it and exits 1 (`resolveBindAddress`/`BindAddressError`, `parseAllowedHosts`/`AllowedHostError`) — so the refusal has the same path and wording as `--bind` (the comparison #689 draws), `options.port` stays a string (commander never runs a parser over a default), and the existing handler test harness covers it.

- **New `src/cli/port.ts`**: `resolvePort` (only an unsigned integer in 0-65535; `PortError` otherwise, message with no trailing period because the handlers append `. The server was not started.`) and `portOption()`, the one flag definition both transports add.
- **Handlers**: `resolvePort` is the first statement inside each `try` (a throw before it would be an unhandled rejection from the commander action), ahead of sse's deprecation line and both `Starting … on port` lines. `createHttpApp`/`createSSEApp` resolve it beside `resolveBindAddress` for the `/health` seed, before `serverFactory` in sse — so nothing is built.
- **Startup lines** move from the `listen()` callback into one `announceListening` helper (`bind-address.ts`) run from the `'listening'` event, after `server.address()` is copied into the `/health` record: they print the bound port with IPv6 hosts bracketed, and the endpoint line is **mirrored to stderr** — both transports silence the console, so the logger alone left a plain `-p 0` with nowhere visible to read the port (stderr is not the MCP channel there; `reportFatal` already uses it). Node registers a `listen()` callback as a `once('listening')` listener *before* ours, so it could only ever repeat the requested port. The `'error'` handler still reports the port the operator typed.
- `AllowedHostError`, `BindAddressError` and `PortError` share a `StartupRefusalError` base; the handlers catch that one class.
- `docs/quickstart.md` flag cheat-sheet gains the range and the `-p 0` rule; `changelog.d/689.fixed.md`.

The dev-proxy resolves `DEV_PROXY_PORT` with the same rule (`resolveBackendPort`, 1-65535): a junk value is refused by name at startup instead of polling `127.0.0.1:NaN/health` for 30 s or truncating `3001x` to 3001.

## Tests

- `tests/unit/cli/port.test.ts` (new): the accepted range, the seven refusal shapes, the option shape.
- `http-command.test.ts` / `sse-command.test.ts`: a refusal test per handler (`abc`, `3001x` → exit 1, `listen`/`serverFactory` never called, stderr names the flag and the value, no startup line announced), a bound-port test per handler (`-p 0` → the log carries 64417 from `server.address()` and never `:0`), and the two existing startup-line tests fire the captured `listening` handler. The listen mocks tolerate the dropped callback (`cb?.()`, two-arg assertions); the base server mocks gain `address()` returning `null` as an unbound `net.Server` does.
- `setup.test.ts`: the description assertions check the range.
- Review pass: `bind-address.test.ts` covers `announceListening` (IPv6 bracketing, no-address fallback, stderr mirror); `port.test.ts` the distinct out-of-range message; `backend-env.test.ts` the dev-proxy guard; the bound-port handler tests assert the stderr line.

Verified locally (Windows): `npm run test:unit` 265 files / 5136 tests pass; `npm run typecheck:all` 0 errors; `npm run lint` clean; `npm run changelog:check` ok; `npm run build` ok; the table above from `dist/index.js`. CI runs no e2e suites, so the live rows are local evidence.

## Self-debug (local evidence)

mcp-debugger stepping through its own CLI via the dev-proxy, `dist/index.js` as the JavaScript debuggee, breakpoints set by `statement:` on the `.ts` sources through the shipped source maps:

1. `args: ['http', '-p', 'abc']`, breakpoint on `throw new PortError(refusal)` (`src/cli/port.ts:54`): stop landed; locals `raw: 'abc'`, `refusal: "--port value 'abc' …"`; stack `resolvePort` ← `handleHttpCommand` (`http-command.ts:473`, the first statement of the `try`) ← commander action ← `main`, no app or server frame; continue → stderr `mcp-debugger: --port value 'abc' … The server was not started.`, session `stopped` with `exitCode: 1`.
2. `args: ['http', '-p', '0', '--log-file', …]`, breakpoint inside the `'listening'` handler: `server.address()` → `{ port: 58485 }`, `listening` already `{ port: 58485 }`, `port` (requested) `0`; continue → the log carries `listening on 127.0.0.1:58485` and `endpoint available at http://127.0.0.1:58485/mcp`, `/health.listening.port` 58485. The debuggee's startup janitor left the live backend's proxy worker alone (owner pid alive). (That run needed `--log-file` to read the line at all — the review's first finding; after the fix a plain `http -p 0` prints `mcp-debugger: MCP endpoint available at http://127.0.0.1:58573/mcp` on stderr, and `--bind ::1 -p 0` prints `http://[::1]:58575/mcp`.)

https://claude.ai/code/session_01PPBa8dFiZU9yPGsCotwvjT


debugmcpdev:
Follow-up commit 41c3b5ad.

## Review adjudication (`/code-review high`: 26 candidates → 17 deduped → 12 survived, 10 reported, 5 refuted)

All ten reported findings were real and are addressed in the follow-up commit.

| # | Finding | Verdict | What changed |
|---|---|---|---|
| 1 | The `-p 0` bound port reached only the winston logger, which has no console transport in http/sse mode — a plain `mcp-debugger http -p 0` printed the port nowhere visible | **Confirmed, the important one** | `announceListening` mirrors the endpoint line to stderr (`mcp-debugger: MCP endpoint available at http://127.0.0.1:58573/mcp`), the path `reportFatal` already uses; stderr is not the MCP channel on either transport. Verified on a plain run with no `--log-file`, http and sse |
| 2 | `callback?.()` loosening had also reached two `mockServer.close` mocks, so the #337 shutdown tests stopped discriminating "exited from the close callback" | Confirmed | Reverted to `callback()` at both sites |
| 3 | `-p 70000` was told it "is not a port number" | Confirmed | Second message: `--port value '70000' is above 65535; give an integer 0-65535 (0 lets the OS choose)` |
| 4 | `${bind.address}:${port}` unbracketed for IPv6 (`::1:64417`) | Confirmed | `authority()` brackets IPv6 for the startup line; `describeEndpoint` shares it. Verified: `--bind ::1 -p 0` → `http://[::1]:58575/mcp` |
| 5 | The dev-proxy side benefit claimed in the PR body was overstated: `DEV_PROXY_PORT=3001x` still truncated and `abc` polled `127.0.0.1:NaN` for 30 s | Plausible → confirmed | `resolveBackendPort` in `backend-env.mjs` (1-65535, unset/empty → 3001), unit-tested; a junk value now fails at proxy startup naming `DEV_PROXY_PORT` and the value |
| 6 | `--port` resolved twice per start (handler + factory) | Accepted | The factory is the single resolve site; the handlers read the port back from the `/health` seed. Side effect stated: the `Starting … on port` line (and sse's deprecation line) now follows the factory, so a `--bind` / `--allowed-host` refusal is no longer preceded by a "Starting" line either |
| 7 | Three `extends Error` refusal classes and a growing `instanceof` chain in two catch arms | Accepted | `StartupRefusalError` base (`src/cli/startup-refusal.ts`); `AllowedHostError`, `BindAddressError`, `PortError` extend it; both handlers catch the one class |
| 8 | The nine-line `'listening'` handler duplicated in both transports | Accepted | `announceListening(server, listening, { logger, proc, label, path })` in `bind-address.ts`, unit-tested directly (IPv6, no-address fallback, stderr mirror) |
| 9 | `expect(mockServerFactory).not.toHaveBeenCalled()` is redundant in the http refusal test | Accepted | Annotated: `createHttpApp` builds servers lazily, the SSE twin's assertion is the load-bearing one |
| 10 | Nine listen mocks per file kept a dead three-parameter shape with `cb?.()` | Accepted | All are `vi.fn(() => server)` now; zero optional-call mocks remain |

Refuted by the verifiers and left alone: the `address()` null fall-through (unreachable inside a `listening` handler), raw-value echo (JSON-escaped in the file transport), the startup janitor concern (owner-liveness based), the factory-only `/health` seed (unchanged from main), the historical header comment (house style of the sibling modules).

Gates after the follow-up: `npm run test:unit` 265 files / 5136 tests pass, `npm run typecheck:all` 0 errors, `npm run lint` clean, `npm run changelog:check` ok, `npm run build` ok.

Issue #687, opened 2026-09-10
javascript launch: consider smartStep: false by default, as attach has since #513 (js-debug resumes skipped-frame pauses through the smart-stepper, not skipFiles)


debugmcpdev:
## Summary

A JavaScript **launch** keeps js-debug's `smartStep: true` by default while **attach** has defaulted it to `false` since #513. js-debug's resume-on-skipped-frame behaviour (the mechanism behind #678) is gated on `smartStep`, not on `skipFiles`: in the vendored `vsDebugServer.js`, `getSmartStepDirection` opens with `if (!this.launchConfig.smartStep || ...) return`, and the blackboxed-frame check lives inside that gate. The script skipper itself only calls `setBlackboxPatterns` and never resumes anything.

PR #686 shipped the opt-in path (`dapLaunchArgs.justMyCode: false` turns the stepper off and un-blackboxes node_modules) plus pending-stop hints that name the mechanism. It did **not** flip the launch default, because that is a product decision with an unmeasured cost. This issue is that decision.

## Evidence (three live runs on the #686 build against `examples/javascript/express_idle_server.js`, recorded by the PR's review)

| Run | Launch | Result |
|---|---|---|
| A | default `skipFiles` (`<node_internals>/**`, `**/node_modules/**`) + `smartStep: false` | `pause_execution` **landed**: `lastStop.reason: "pause"`, frame 0 `<node_internals>/internal/timers:527` |
| B | `skipFiles: ["<node_internals>/**"]` + `smartStep: true` (`justMyCode: false` with the stepper forced back on) | the #678 chase reproduced with node_modules **not** blackboxed: 323 `Debugger.paused` / 646 continued / 0 `stopped`, still `running` 90s later |
| C | default `skipFiles` + `smartStep: false`, `step_over` from a breakpoint inside express | `"Stepped over"`, `state: "paused"` at `<node_internals>/events:513`, not pending (V8's own blackboxing turns the step into a step-out; landing on the next *dependency* line still needs node_modules off the list, which `justMyCode: false` does) |

So with `smartStep: false` and the skip list unchanged, a pause on a launched server lands (in an internals frame, marked as such by the stack response) and a step from a dependency frame lands, while node_modules and internals stay hidden from display and from step-into.

## The open question

Should launch default to `smartStep: false`, as attach does?

[...]

Issue #513, opened 2026-08-27
js attach: pause_execution stays pending forever on an idle Node server — the stop never lands even once JS runs


debugmcpdev:
## Summary

On a **javascript attach** session against an idle Node HTTP server (an mcp-debugger instance, as it happens), `pause_execution` returns the documented `pending: true` shape — but the promised follow-up stop **never lands**, even after the target demonstrably executes plenty of JavaScript (it served multiple HTTP requests in the interim). The session stays `running` indefinitely; `get_stack_trace` keeps answering "Session is not paused". Retried twice minutes apart, same result.

## Repro

1. `node --inspect=127.0.0.1:9339 dist/index.js http -p 3111` (any long-lived mostly-idle Node server works)
2. `create_debug_session {language: javascript}` → `attach_to_process {port: 9339, stopOnEntry: false}` — succeeds, state `running`
3. `pause_execution {sessionId}` → `{"success":true,"state":"running","data":{"pending":true,"message":"Pause requested; no 'stopped' event within 5s …"}}`
4. `curl http://127.0.0.1:3111/health` several times — the server responds 200 each time, so JS is executing
5. `list_debug_sessions` → still `running`; `get_stack_trace` → `"Session is not paused (state: running)"` — held for minutes

## Expected

The tool's own contract: "The session will report 'paused' once the stop lands." With the event loop provably running JS, the stop should land on the next dispatch — this is how CDP `Debugger.pause` behaves under an IDE.

## Notes / suspicions

- Likely the pause is issued against the wrong thread/target in the js-debug parent/child topology (attach adopts a child session for the real target; `threadId 0` quirks in that path have bitten before — cf. #275/#352/#353 for other adapters' variants of "pause doesn't actually pause").
- **Workaround that works every time**: set a real breakpoint on a line the server executes per request (`set_breakpoint` on the /health handler) and trigger it — lands immediately with a full stack. This is what the #512 investigation used to get an `evaluate_expression` context.
- Found while dogfooding #501/#512 (attaching mcp-debugger to a live mcp-debugger server); it cost the investigation an instrumentation window and would confuse any agent following the tool's guidance.

## Environment

main @ 9b220a20, Linux, Node 22.22.3, vendored js-debug.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NgA5kHzkhFZLqbWNZRTsVG

PR #702, opened 2026-09-11
test(javascript): e2e for the unset-default launch, README states why the stepper stays on (#687)

debugmcpdev:
Refs #687 (measured and kept — the matrix and verdict are in [the issue comment](https://github.com/debugmcp/mcp-debugger/issues/687#issuecomment-5637908662)).

Tests and docs only; no runtime change.

## What changed

- **e2e** (`tests/e2e/mcp-server-smoke-javascript.test.ts`): every #678 test passed `justMyCode` explicitly, so the README's *default* row was untested end-to-end. `launchIdleExpress` / `pauseInsideExpress` now take a launch object (`{ justMyCode?, adapterLaunchConfig? }`) and spread `justMyCode` only when given. Two new tests:
  - the **unset default** launch: a `step_over` from the express breakpoint stays `pending: true` with the skipped-frame hint and the `justMyCode: false` remedy;
  - the **stepper-off launch on the default skip list** (`adapterLaunchConfig: { smartStep: false }`, the post-flip shape #687 proposed): the step lands, in an internals frame the stack response marks (`pausedFrame`), with readable locals.
- **`docs/javascript/README.md`**: after the stepping table, why launch keeps the stepper on while attach (#513) does not — the measured costs of turning it off: `step_into` from a dependency frame stops in Node internals one line at a time (`<node_internals>/**` on `skipFiles` is not a V8 blackbox), and `step_into` an `async` callee in ES2015-or-lower TypeScript costs four unmapped `__awaiter` stops; `step_over` is unchanged.

## Verified locally (Windows)

`npx vitest run tests/e2e/mcp-server-smoke-javascript.test.ts` → 9/9 (the two new tests 12.1 s and 7.2 s); `npm run typecheck:tests` 0 errors; `npm run check:docs` clean. CI runs no e2e suites, so the run above is the evidence.

https://claude.ai/code/session_01PPBa8dFiZU9yPGsCotwvjT


debugmcpdev:
Follow-up commit 617f74c9.

## Review adjudication (`/code-review medium`: 8 findings, all addressed in the follow-up commit)

| # | Finding | What changed |
|---|---|---|
| 1 | "`<node_internals>/**` on `skipFiles` is not a V8 blackbox" is wrong as a mechanism: js-debug pushes `^(node:)?internal\/` plus one pattern per builtin, but builds the builtin set as `builtinModules.map(i => i + '.js')`, so `node:url` / `node:events` escape while `node:internal/*` is blackboxed | Verified in the vendored bundle (`_initNodeInternals`, `_updateGeneratedSkippedSources`). README states that mechanism; the e2e comment too. The #687 issue comment carries a correction |
| 2 | "Attach sessions default neither key" is stale for `smartStep` (#513) | README now says attach defaults `smartStep` to `false` and leaves `skipFiles` unset. The same sentence in `adapter-policy-js.ts:245` is a source comment and is left for the next `packages/`-touching PR (a changelog fragment for a comment edit is not worth it) |
| 3 | The intro overgeneralized "neither a pause nor a step issued from a skipped frame lands" | Qualified to `step_over` / `step_out`; `step_into` still reaches the next call into user code |
| 4 | `expect(stack.pausedFrame).toBeDefined()` measured async-ancestor availability | Asserts `stackFrames[0].file` matches `<node_internals>`, the stop frame in both resolver branches |
| 5 | The unset-default `step_over` test is byte-identical below the MCP boundary to the `justMyCode: true` one | Folded into an `it.each` over `[{ justMyCode: true }, {}]`; the helper comment says `{}` is a surface-level replay (the server merges `defaultDapLaunchArgs`) |
| 6 | "ES2015 or lower" excludes ES2016, which also emits `__awaiter` | "a target below ES2017" |
| 7 | The load-bearing `step_into` claim had no test in either direction | `it.each`: stepper on → `express_idle_server.js` in one press; `adapterLaunchConfig: { smartStep: false }` → `<node_internals>/url`. Both pass locally (7.1 s each); this is the reopen tripwire for #687 |
| 8 | Session name labelled the `smartStep: false` launch `js-dependency-default` | `describeLaunch` encodes `justMyCode` and every `adapterLaunchConfig` key |

Verified: `npx vitest run tests/e2e/mcp-server-smoke-javascript.test.ts` 11/11; `npm run typecheck:tests` 0; `npm run check:docs` clean.

